PR #3705, opened 2017-12-08
Update tools/pedestrianFlow.py

kubasevca:
Correction of variable name 'fouttrips' (not known) -> 'f'


namdre:
thanks

behrisch:
@kubasevca We want to add GPLv2+ as a secondary license to sumo and need your permission to also include this (and maybe code from other pull requests) in this license change. If you agree please add a +1 to #6506. This ticket also has more details and if you have questions you can ask them there. Thank you in advance!

PR #3713, opened 2017-12-14
Place turn defs within an interval element, fixes #3712

### LINKED ISSUE #3712: generateTurnDefs.py creates turn definition file with incorrect schema
While creating the turns XML file, generateTurnDefs.py does not place the generated turn definitions within an interval element, as required by the [turns file schema](http://sumo.dlr.de/xsd/turns_file.xsd).

srishtidh:
Hello,

Here are the changes I've made to ensure generateTurnDefs.py generates turn defs files with the correct schema:

- Update turndefinitions.to_xml() to place the turn definitions within an interval element
- Add interval begin (default 0) and end (default 3600) times as command-line options to generateTurnDefs.py
- Update the generateTurnDefs/uniform_distribution test accordingly (both the options and the expected turn definitions)

Does this seem all right?

### DIFF
--- a/tests/tools/turndefs/generateTurnDefs/uniform_distribution/options.tools
+++ b/tests/tools/turndefs/generateTurnDefs/uniform_distribution/options.tools
@@ -1,1 +1,1 @@
---connections-file connections.con.xml --turn-definitions-file turn-defs.xml
+--connections-file connections.con.xml --turn-definitions-file turn-defs.xml --begin 100 --end 200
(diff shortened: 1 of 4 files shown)


namdre:
Thanks!

behrisch:
@srishtidh We want to add GPLv2+ as a secondary license to sumo and need your permission to also include this (and maybe code from other pull requests) in this license change. If you agree please add a +1 to #6506. This ticket also has more details and if you have questions you can ask them there. Thank you in advance!

srishtidh:
@behrisch Done!

Issue #3099, opened 2017-05-12
retrieve collision information (trac #3099)


behrisch:
traci.simulation.getCollisionNumber
traci.simulation.getCollisionIDList

also a function to determine collider/victim for each collision?

Migrated from http://sumo.dlr.de/ticket/3099
```json
{
    "status": "new", 
    "changetime": "2017-05-12T06:52:55Z", 
    "description": "traci.simulation.getCollisionNumber\ntraci.simulation.getCollisionIDList\n\nalso a function to determine collider/victim for each collision?", 
    "reporter": "namdre", 
    "cc": "", 
    "resolution": "", 
    "_ts": "1494571975864012", 
    "component": "TraCI", 
    "summary": "retrieve collision information", 
    "priority": "major", 
    "keywords": "conference2017", 
    "time": "2017-05-12T06:52:55Z", 
    "milestone": "1.1.0", 
    "owner": "", 
    "type": "enhancement"
}
```

PR #3747, opened 2017-12-31
add traci functionality to get collision count

c433li:
It's useful to know whether collisions have occurred through traci when training self-driving cars under SUMO.


namdre:
Thanks for the contribution. However, it has been rejected in favour of [f871c8e99f7] which is better aligned with the rest of the design and should also supply the desired capabilities (see #3099).

PR #3778, opened 2018-01-24
Added setSpeedMode to c++API

mrtzz:
I needed the already defined "setSpeedMode" Function, so i added it. maybe it's useful to someone.


behrisch:
@guetlein We want to add GPLv2+ as a secondary license to sumo and need your permission to also include this (and maybe code from other pull requests) in this license change. If you agree please add a +1 to #6506. This ticket also has more details and if you have questions you can ask them there. Thank you in advance!

PR #3782, opened 2018-01-27
add missing traffic light getters to TraCI C++ API

riebl:


behrisch:
Thanks for the patch! Could you please also add a test call to this function https://github.com/DLR-TS/sumo/blob/f19c50d12d4e2db284312886dd507665696a3ebc/src/traci_testclient/TraCITestClient.cpp#L691

riebl:
Done.

behrisch:
Thank You!

PR #3889, opened 2018-03-10
Bugfix for Visual Studio 2015 (and others)

seb-ehm:
As mentioned in #3565, Visual Studio 2015 fails to compile. However, the fix proposed in #3565 did not address the root of the issue. The reason for the compiler errors are macro definitions by Fox Toolkit, which can be disabled.

`fxdefs.h` of the Fox Toolkit defines several macros for float functions (e.g. powf(x,y), line 644 and following ) that, according to the C++ standard, are to be defined in `<math.h>`.
As `fx.h` includes `math.h` before it includes `fxdefs.h`, this does not immediately cause problems with some compilers. But for Visual Studio 2015, `<cmath>` uses e.g. `powf` to calculate `pow(float x, float y)`. Thus, if `<cmath>` is included after `fx.h` (potentially by some other standard header like `<iostream>`), the compilation fails, as the macro conflicts with the usage of powf in `<cmath>`. For GCC, `<cmath>` uses `__builtin_powf(__x, --y);` to calculate `pow(float x, float y)`, which is not affected by the macros in `fxdefs.h`.

A minimum example that compiles with GCC, but fails with Visual Studio 2015 is:

    #include <math.h>
    #define powf(x,y)   ((float)pow((double)(x),(double)(y))) // #include <fx.h>
    #include <cmath>
    
    int main(){}
Adding -DFLOAT_MATH_FUNCTIONS to the compile flags disables the definition of the macros in `fxdefs.h` and thus avoids the errors mentioned above. In my opinion, it should be added for all compilers, as the macros are unnecessary and could cause more problems in the future.


behrisch:
@palvarezlopez could you please have a look. Did this come up in your build tests?

palvarezlopez:
Ok, I check it.

palvarezlopez:
@seb-ehm your solution works as expected, thanks!

seb-ehm:
No problem at all.

Issue #3836, opened 2018-02-20
Build Errors


ab2229:
I am currently attempting to compile SUMO on Windows using Microsoft Visual Studio 2017 with CMake. I have followed the manually generated tutorial given on the [Wiki](http://sumo.dlr.de/wiki/Installing/Windows_CMake), as the automatic generation did not produce any output.

I have pulled the most recent [libraries](https://github.com/palvarezlopez/SUMOLibraries) and ensured they are all in my path. 

Unfortunatly, each time I attempt to compile the project I am given the error message:

MSB6006 "cmd.exe" exited with code 9009. Project - gl2ps. File - Microsoft.CppCommon.targets. Line - 171

I would be very grateful if anyone would be able to assist me with this problem.

Thank you,

Aaron

behrisch:
Do you have git installed and is it in your path?

ab2229:
Yes, I do

seb-ehm:
@ab2229 I just added a commit that should fix your issue.

PR #3910, opened 2018-03-18
Fixed a copy command (re #3836)

seb-ehm:
The build error in #3836 is caused by the command `cp`, which supposedly only works on Windows when cygwin is installed . The usual command to copy files on Windows is `copy`. However, `copy` does not work properly when paths contain slashes. I therefore replaced the copy command with the CMake built-in copy command. I also added `$<CONFIG>` to the path, as otherwise the command will fail if Sumo isn't compiled in Release mode.

For the same reason, I added the path to the Debug build in the following `FIND_LIBRARY`call.

However, there is another issue with the inclusion of GL2PS on Windows, as `HAVE_GL2PS` is only going to be set to 1 after manually re-running CMake after an initial build of GL2PS. But I think that warrants a new separate issue.


behrisch:
Thanks a lot!

PR #3954, opened 2018-03-28
Fix memory leaks

seb-ehm:
Cppcheck discovered some mismatching allocations and deallocations (delete / delete[]) and memory leaks.


namdre:
thank you

PR #3970, opened 2018-04-06
sumolib net reading now parses optional lane width attribute

gprichter:
When checking the output of a conversion from OpenDrive files
missing lane data was required.
The optional attribute `width` in the XML `lane` element was needed,
but not parsed by the `sumolib.net`, previously 

**NOTE:** a default argument value in the `Lane.__init__()` might be wise to better maintain backwards compatibility, but was omitted to stay in line with the other optional attributes' arguments.


namdre:
duplicate of https://github.com/DLR-TS/sumo/issues/3842 (already implemented)

PR #3983, opened 2018-04-11
traci: add VehicleScope::setSpeedFactor in C++ API

riebl:
First pull request at new location? ;-)

### DIFF
--- a/src/traci_testclient/TraCITestClient.cpp
+++ b/src/traci_testclient/TraCITestClient.cpp
@@ -830,6 +830,7 @@ TraCITestClient::testAPI() {
     std::pair<std::string, double> leader = vehicle.getLeader("1", 1000);
     answerLog << "    getLeader: " << leader.first << ", " << leader.second << "\n";
     vehicle.rerouteTraveltime("0");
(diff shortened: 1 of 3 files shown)


behrisch:
Thanks for the contribution! Now that we are in Eclipse we need you to open an Eclipse account and complete a small form confirming that you agree with the licensing under EPL v2 before we can accept it. Please see http://sumo.dlr.de/wiki/FAQ#How_do_code_contributions_work.3F. You only need to do this once.

behrisch:
Thanks for your fast response. Unfortunately there seems to be another requirement which we did not see yet. Your commit needs to be signed off. You can do this locally by issuing "git commit --amend -s" but this will modify the commit so you will need to do a force push afterwards. Sorry for the hassle but as you noticed it is the first time :-)

riebl:
I have always wondered who is actually using those signed-off markers, now I know ;-)
@behrisch Oops, I have moved the command invocation a few lines up by now.

behrisch:
Thanks for the contribution and for working this through with us!

PR #3995, opened 2018-04-16
traci: add getSignals and setSignals to VehicleScope

riebl:
VehicleScope::getSignalStates lacked the implementation. It has been replaced by VehicleScope::getSignals which has the same method name as its Python API equivalent.


namdre:
Thanks!

PR #4067, opened 2018-05-14
Missing GNEParkingAreaReroute.h/.cpp in MSVC netedit_additionals project

m-kro:
For me, Netedit failed to compile showing Linker errors with respect to GNEParkingAreaReroute. That is why I added them to the netedit_additionals project. 
A short scan of all MSVC project files shows that they haven't been referenced elsewhere either.


palvarezlopez:
@m-kro Hi, Thanks for the patch. The reason is, that MSVC projects will not be supported in the next SUMO Versionen. Instead, we'll use CMake.

m-kro:
Sorry for that. Sometimes I still do not really know what I am doing when using git :-( Actually I wanted to seperate them because the second commit has nothing to do with the first message of the pull request. But I will reopen the request so that you can merge them.

Thanks

behrisch:
Thank you for contributing. but please be aware that the visual studio projects are going to be removed soon.

PR #4152, opened 2018-06-03
Setuptools

bstriner:
Hi everybody!

A few suggested updates for cmake and python libsumo bindings. Just tested this out on windows and linux.
- Use PYTHON_EXECUTABLE instead of assuming that .py files run correctly. Fixes some build issues on my windows machine.
- Add flags to disable building python and jni bindings, instead of always doing it if available
- Restructure python bindings to install with setuptools. This requires staging all of the files and running `setup.py`.
- Add a target (`install_pylibsumo`) if all you want to do is install the python bindings.

This is correctly working for now, but there are a few changes to make:
- Since there is now a setup.py, should fill out the author, licence, url, etc. 
- Version isn't being set in the python module. Need to repurpose the generate-version-h code to get that part working correctly.

BTW, I haven't tested the Mac build.

Cheers


bstriner:
https://github.com/eclipse/sumo/pull/4154

PR #4159, opened 2018-06-05
Close function for libsumo 

bstriner:
Function for closing libsumo. Otherwise, no way to run multiple iterations without getting "A network was already constructed".

This runs but should be tested for any memory leaks on multiple creates/deletes. I think those 4 lines of cleanup are enough but please let me know if there is anything obvious that is missing.

A friend of mine will be running some RL using this code shortly, so that will be some additional testing.

Cheers

### DIFF
--- a/src/libsumo/Simulation.cpp
+++ b/src/libsumo/Simulation.cpp
@@ -25,6 +25,7 @@
 #include <utils/common/StdDefs.h>
 #include <utils/common/StringTokenizer.h>
 #include <utils/common/StringUtils.h>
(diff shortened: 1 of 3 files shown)


behrisch:
Thanks a lot.

behrisch:
@bstriner  We want to add GPLv2+ as a secondary license to sumo and need your permission to also include this (and maybe code from other pull requests) in this license change. If you agree please add a +1 to #6506. This ticket also has more details and if you have questions you can ask them there. Thank you in advance!

Issue #3850, opened 2018-02-27
Add new mode for editing right of way at priority junctions


namdre:
* hotkey 'w'
* similar to connection mode:
 * when entering mode, enable show-connections
 * color legend as described below 
 * first click selects a connection
   * connections that intersect are highlighted:
    * dark green for those which yield to the selected
    * red for those with priority over the selected
 * second click on any intersecting connection toggles right-of-way
 * buttons ok/cancel
 * button 'reset' (resets to default right of way)

Currently, right-of-way can only be configured via xml inputs with a mix of 
* edge priority
* connection attribute 'pass'
* prohibitions 

It is probably not possible to map arbitrary right of way rules to the above 3 mechanisms. In that case, there would need to be extensions to the plain-xml format and to the way right-of-way logic is handled in NBRequest.

The first step is to lay the netconvert groundwork (Jakob)
The second step is to implement the new mode in netedit (Pablo)

namdre:
@palvarezlopez This is not urgent. I'm just putting in here to get it out of my head.

m-kro:
I would second this prohibition mode. Sometimes, the guessed yielding is just not what you want. I started coding on a prohibition mode before watching the issue list. For now, it consists only of the color legend as described above (see screenshot):
![prohibitionmode](https://user-images.githubusercontent.com/33294728/41248897-ab567a56-6db2-11e8-9e90-7d29cb74a5ec.png)
Is there a way to collaborate and speed up this topic?

namdre:
I think making the right of way rules visible is good first step and a worthwhile feature all by itself.
If you send a pull request that adds this mode i'll be happy to review and merge it.
The next step is to figure out to which extent the internal structures and the xml input formats need to be extended.
For that it would help if you could provide a few examples where the default right of way rules are incorrect (the original .net.xml file and a snippet of the <junction> element as you would prefer it to be).

namdre:
refinements are need because the currently used methods 
node->foes() and node->forbids() do not always reflect the final state of the junctionLogic (see the additional checks in NBRequest::getResponseString and NBRequest::getFoesString).

Furthermore the methods would currently retrieve invalid results for lefthand networks due to the way they are computed (via mirroring twice). A possible solution would be to use the cached logic matrix that is proposed as a fix for #4256

namdre:
* Right of way mode is now consistent with .net.xml
* Right of way can be customized in detail by setting rightOfWay=edgePriority and then customizing edgePriorities. 

[...]

Issue #4212, opened 2018-06-12
refactoring initialization of attribute position in list


namdre:
issue #4211 demonstrated that the current method is error prone.

namdre:
the referenced commit is my suggestion for refactoring. maybe you have a better idea.

PR #4231, opened 2018-06-14
Netedit prohibition (visualization) mode

m-kro:
This pull request addresses issue #3850 partially. A new mode is added to Netedit which allows for visualizing the give way state of connections. First, a connection is selected with LMC and all other conflicting are coloured according to their state:
- undefined (there is a conflict but no give way rule): grey
- yielding: green
- has give way: red
Info about the selected connection is displayed in the left frame. A different connection can be selected after clicking "Cancel".


namdre:
The current merge-build fails due to an error in GNEProhibitionFrame.h
(https://travis-ci.org/eclipse/sumo/jobs/392329849). Could you check this? Possibly it came about due to recent refactoring (#4212)

Issue #4318, opened 2018-07-10
Update Car Following Model "Rail" to support moving-block operation


palvarezlopez:
currently, the function MSCFModel_Rail::followSpeed completely ignores the leader vehicle

Assigned to: https://github.com/FlL-DLR

PR #4372, opened 2018-07-31
Implemented driving in absolute braking distance (including a safety …

palvarezlopez:
…gap) for railways.

Refs #4318


namdre:
@palvarezlopez @FlL-DLR can you amend the commit with the proper signed-of information: 

> 
> Make sure all the commits you want to contribute have a valid Signed-off-by entry. If you use command line git, this is as simple as doing "git commit -s" when committing.
> 
> if you want to modify an existing commit because you forgot to sign it, you can use "git commit -s --amend"

behrisch:
Except for the IP check the code looks fine. Concerning the file header: The date should not be updated but stay at the date of initial creation.

PR #4466, opened 2018-08-28
Introducing fare modul for mdv's fare system

reuler:


behrisch:
We will still need to refactor this in using params instead of new attributes, but first we need to get this working

PR #4663, opened 2018-10-03
Fix function argument names and order

seb-ehm:
The argument names and order for setParameter(...) differs between function declaration (name, value, string) and definition (id, name, value). This can lead to user errors and is fixed by this commit.


namdre:
Thanks for pointing out the issue. I've made the argument names conform to the style used by other domains.

Issue #4670, opened 2018-10-04
Memory Leak in AStarRouter


seb-ehm:
The member variable [myLookupTable](https://github.com/eclipse/sumo/blob/65b48fbf1c31216e3bb61a839c7b9a36bb952702/src/utils/vehicle/AStarRouter.h#L343) in AStarRouter.h is a pointer to memory owned by the router, but allocated in external code.
Both in MSDevice_Routing.cpp ( [e.g. here](https://github.com/eclipse/sumo/blob/65b48fbf1c31216e3bb61a839c7b9a36bb952702/src/microsim/devices/MSDevice_Routing.cpp#L419)) and in duarouter_main.cpp ( [e.g. here](https://github.com/eclipse/sumo/blob/65b48fbf1c31216e3bb61a839c7b9a36bb952702/src/duarouter/duarouter_main.cpp#L115)), memory is allocated for the LookupTable, but will never be freed even after destruction of the router.

seb-ehm:
I made a pull request with a suggestion on how to fix the potential memory leak. But there might be a more elegant solution - or the potential memory leak might even be a non-issue.

PR #4671, opened 2018-10-04
Fix AStarRouter Memory Leak

seb-ehm:
The member variable myLookupTable is a pointer to memory owned by AStarRouter.
However, the memory currently is never freed (#4670, #13). As the AStarRouter can be
[cloned](https://github.com/eclipse/sumo/blob/65b48fbf1c31216e3bb61a839c7b9a36bb952702/src/utils/vehicle/AStarRouter.h#L163), a simple delete in the destructor could lead to use-after-free
errors. I therefore modified myLookupTable to be a shared pointer, which
should allow cloning but free the memory after the last copy/clone of the
router is out of scope.

### DIFF
--- a/src/duarouter/duarouter_main.cpp
+++ b/src/duarouter/duarouter_main.cpp
@@ -31,6 +31,7 @@
 #include <string>
 #include <limits.h>
 #include <ctime>
+#include <memory>
 #include <xercesc/sax/SAXException.hpp>
 #include <xercesc/sax/SAXParseException.hpp>
 #include <utils/common/TplConvert.h>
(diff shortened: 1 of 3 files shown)


behrisch:
Thanks for the patch. I just noticed that your commit is not signed and we probably do not have an Eclipse CLA from you. Could you please create an Eclipse account (if you don't have one already) and fill out the form as described here: http://sumo.dlr.de/wiki/FAQ#How_do_code_contributions_work.3F

behrisch:
Thanks alot.

seb-ehm:
I previously had some issues receiving the confirmation e-mail for my Eclipse account, thank you for reminding me to create one.

Issue #4699, opened 2018-10-11
Memory Leak in netload/NLTriggerBuilder.cpp


seb-ehm:
Cppcheck identified several memory leaks in [NLTriggerBuilder.cpp](https://github.com/eclipse/sumo/blob/d6aa3c9859ec41d84b54700c9d2f513444ffa860/src/netload/NLTriggerBuilder.cpp), as memory is allocated via several Build* methods and owned by pointers named `trigger`. When the condition `if (file == "")` is not fulfilled, the memory is not freed, which leads to memory leaks in lines [143](https://github.com/eclipse/sumo/blob/d6aa3c9859ec41d84b54700c9d2f513444ffa860/src/netload/NLTriggerBuilder.cpp#L143), [334](https://github.com/eclipse/sumo/blob/d6aa3c9859ec41d84b54700c9d2f513444ffa860/src/netload/NLTriggerBuilder.cpp#L334), [339](https://github.com/eclipse/sumo/blob/d6aa3c9859ec41d84b54700c9d2f513444ffa860/src/netload/NLTriggerBuilder.cpp#L339) and [385](https://github.com/eclipse/sumo/blob/d6aa3c9859ec41d84b54700c9d2f513444ffa860/src/netload/NLTriggerBuilder.cpp#385).

I've prepared a pull request that adds some delete calls at the appropriate places, but this does not completely resolve the issue, and the memory owned by trigger still leaks.

**Mem Leak 1**
There are several try-catch-blocks catching ProcessErrors, which suggests that some of the functions used can throw. I am not sure whether the triggers are still being used after error handling, but it might be appropriate to `delete trigger` in the catch blocks. Using a smart pointer might be a good choice, but that would require significant changes:

**Mem Leak 2?**
I tried to follow the owners of trigger. This is going to be a bit lengthy and I probably spent too much time on it. Nevertheless:

 All of the triggers inherit from GenericSaxHandler and use its [register parent method](https://github.com/eclipse/sumo/blob/d6aa3c9859ec41d84b54700c9d2f513444ffa860/src/utils/xml/GenericSAXHandler.cpp#L162) 
https://github.com/eclipse/sumo/blob/d6aa3c9859ec41d84b54700c9d2f513444ffa860/src/netload/NLTriggerBuilder.cpp#L139-L142
 The trigger keeps a reference to itself in its member variable myParentHandler and forwards itself to `XMLSubSys::setHandler`:
[...]

PR #4700, opened 2018-10-11
Fix some memory leaks in NLTriggerBuilder (Re #4699)

seb-ehm:
At several places in the file, memory is allocated and owned by
a pointer named trigger. If the condition "file == "" " is not
fulfilled, a memory leak occurs, as trigger is not deleted. I've added
the appropriate delete calls for these cases.
I suspect that even in the regular case (file==""), the memory also
leaks, as I could not identify any corresponding delete calls in the
parents/owners of trigger (see issue #4699).

Signed-off-by: Sebastian Ehmann <sebastian@dipolmoment.de>


seb-ehm:
Ok, I definitely underestimated the side effects of the constructor of the trigger. It registers itself in MSEventControl and is deleted with the destructor of MSEventControl.

PR #4702, opened 2018-10-12
New lisum.xml structure

maxidigital:


behrisch:
@maxidigital  We want to add GPLv2+ as a secondary license to sumo and need your permission to also include your code contributions in this license change. If you agree please add a +1 to #6506. This ticket also has more details and if you have questions you can ask them there. Thank you in advance!

PR #4712, opened 2018-10-16
Fix null reference in GNELane.cpp

seb-ehm:
Creating a null reference by dereferencing a null pointer is undefined
behaviour. This fix replaces the null reference with a minimal GNEEdge
instance.

Signed-off-by: Sebastian Ehmann <sebastian@dipolmoment.de>


seb-ehm:
Should have tested it in GCC, not only in VS2017. Passing an `rvalue` as a non-const reference might be tolerated by Visual Studio, but is against the standard as well. As long as `myParentEdge` is a reference, I can't think of a standard conforming way to initialize it in a default constructor. So it likely would be better to change `myParentEdge` to a pointer.

namdre:
One solution would be to instantiate a single static (dummy) GNEEdge which supplies the needed reference

seb-ehm:
Interesting suggestion, I'll have a go at it when time permits.

PR #4720, opened 2018-10-18
Clang automatic fixes

seb-ehm:
First of all, this is quite a large pull request with no functional changes. It is all about refactoring. If the SUMO core team prefers smaller pull request (or if this kind of refactoring is unwanted in general), just let me know.

I've used clang-tidy in combination with clang-apply-replacements to generate and apply fixes for two changes:
- Use `nullptr` instead of `0` or `NULL`
The reasoning behind this kind of change can be best expressed by source code:
A certain line in MSEdge.cpp currently looks like this:
https://github.com/eclipse/sumo/blob/dfa43d556844ae46a3dffbcc12fe9772633dab05/src/microsim/MSEdge.cpp#L154
And is going to be refactored to this:
https://github.com/eclipse/sumo/blob/271284e3e5a4c7bc026221f9c196fe26e21bf369/src/microsim/MSEdge.cpp#L154
Before the refactoring, it looks as if an array would be accessed. But what actually happens is that a `nullptr` is used / inserted as a key in a map. Whether this is the intended behaviour or not can only be questioned when it is obvious that a pointer is used here.

- Use range-based for loops
This change is supported by the SUMO developer guide lines which state that range-based for loops are to be preferred (whether they should be refactored at once is a different question; as state in my introduction, just let me know if this is not desired).
I've separated this change in three commits. The first commit shows all replacements by clang-apply-replacements. However, the fixes are not perfect and not all files compile. Mainly, because at several places, clang missed to add a space when a `delete` is immediately followed by a `(`. For example, this:
https://github.com/eclipse/sumo/blob/dfa43d556844ae46a3dffbcc12fe9772633dab05/src/guisim/GUINet.cpp#L91-L93
has been automatically refactored to this:
https://github.com/eclipse/sumo/blob/784bc8bb776b475be8af547d0e217966d5d09d7e/src/guisim/GUINet.cpp#L91-L93
These kind of errors (and one wrong template replacement and some variable names) are fixed by commit https://github.com/eclipse/sumo/commit/784bc8bb776b475be8af547d0e217966d5d09d7e , which has been created manually.
The third commit https://github.com/eclipse/sumo/commit/ed6aac54f416bf93971aca609681997cffcea5c2 adds some more `auto`s (added manually as appropriate), as clang somehow missed a few opportunities to use them, e.g. here:
https://github.com/eclipse/sumo/blob/784bc8bb776b475be8af547d0e217966d5d09d7e/src/gui/dialogs/GUIDialog_Breakpoints.cpp#L179
A nice thing about clang-tidy fixes is that it automatically tries to guess variable names. This:
https://github.com/eclipse/sumo/blob/dfa43d556844ae46a3dffbcc12fe9772633dab05/src/microsim/MSTransportableControl.cpp#L52-L55
is changed to this (I manually fixed the missing space after `delete` here):
https://github.com/eclipse/sumo/blob/aa3c6c66793ecf1f2ef5d4fec78540e981a10dea/src/microsim/MSTransportableControl.cpp#L52-L55
just by removing the trailing `s`. This obviously does not work everywhere, so I manually changed some variable names from, for example `for (const auto & mySpaceOccupancie : mySpaceOccupancies)` to `for (const auto & mySpaceOccupancy : mySpaceOccupancies)` when I noticed them.

Obviously, a lot of files and lines of source code have been modified. If it is too hard to review these kind of changes, again, just let me know.

Now, I'm keeping my fingers crossed that I haven't missed any errors and that AppVeyor and Travis are happy.


namdre:
Thanks for the effort!
- The 0 -> nullptr patch is pretty straight forward. If you could seperate this into it's own pull request I'd be happy to merge it now.

- the conversion to range based loops carries a slight risk of semantic corrutipn since iterators are occasionaly used outside the loop. Also, some patterns such as erasing from a vector still require iterators. I'm not familiar enough with the clang automation to merge this without extensive review. I've discussed this with my collegue and we would prefer to do this for each code directory seperately (netbuild, netimport, router, ....)

- there is an additional concern regarding the use of auto. We've discussed this among the developers and decided that we want to keep typed loop veriables whenever this would enhance readability of the code (a trade-off between verbosity and hiding variable types). This means we'd mostly use auto whenenver the type is unwieldy (map iterators) but still write things like for 
`(for NBEdge* e : someVector)`. In automatic conversion i'd forgo the 'auto' replacement completely. And always use the fully qualified type.

- regarding the name of the loop variable, I'd prefer keeping the name of the iterator as in your replacement example
`for (auto & i1 : myJunctionWrapper) { `
This is because the iterator name is succinct and was occasionally chosen to convey some information whereas using the container name without the 's' suffix is needlessly verbose.

- we would prefer to attempt loop refactoring in december (when there is more time to do code review). To make the code review managable we would prefer to do this for a single subfolder first.

- there seems to be a merge conflict right now anyway due to continous work on the affected sources

so in short:
- please seperate the nullptr patch into it's own pull request
- please do another pullrequest with minimal range-based-loop replacement (no auto, keep the old iterator variable name) for only the src/netwrite folder. If this works well we could then attempt other folders

regards,
Jakob

seb-ehm:
Thank you for your response. You address some valid points. I am going to make a separate pull request for the `nullptr` commit.
Regarding the loop conversion: as far as I can tell, clang-tidy currently has no option to avoid replacing the type of the loop variable with `auto`.
clang-tidy does some more checks for the name of the variable, for example
https://github.com/eclipse/sumo/blob/c08987cc2689c7a00bd0805316e084d1b852ae56/src/dfrouter/RODFDetector.cpp#L102-L103
is replaced by
https://github.com/eclipse/sumo/blob/784bc8bb776b475be8af547d0e217966d5d09d7e/src/dfrouter/RODFDetector.cpp#L102
but you are right that, in general, variable renaming can lead to verbose and unwanted changes. There does not seem to be an option to avoid renaming the loop variable.

It would be nice to have an interactive refactoring session, e.g. to chose the variable name, but that's currently [not possible:
](https://github.com/llvm-mirror/clang-tools-extra/blob/57b58fc65c27b88bfd9642ee17164d86673bd4a8/clang-tidy/modernize/LoopConvertUtils.cpp#L844-L847)
```
 // FIXME: Add in naming conventions to handle:
 //  - How to handle conflicts.
//  - An interactive process for naming.
```

Regarding the usage of iterators outside of a loop, clang-tidy does seem to check whether the identifier [has been declared in a parent context](https://github.com/llvm-mirror/clang-tools-extra/blob/57b58fc65c27b88bfd9642ee17164d86673bd4a8/clang-tidy/modernize/LoopConvertUtils.cpp#L889), but I am also not familiar with the inner workings of clang-tidy and am not sure whether all potential error cases are handled correctly. So I can totally understand your preference to apply fixes one-by-one in smaller parts of the codebase.

I tried to find some additional, external tooling for clang-tidy that would allow for a more detailed refactoring (e.g. naming of loop variables), but had no luck so far. But if I find something, I'll have a look at `src/netwrite` and let you know. Thank you again for your detailed answer.

Issue #4850, opened 2018-11-26
clang error: ld: unknown option: --start-group


schwamborn:
Building from source with cmake under macOS High Sierra by running

    cmake ../..
    make

results in the following error:

```
Scanning dependencies of target testcommon
[ 99%] Building CXX object unittest/src/utils/common/CMakeFiles/testcommon.dir/StringTokenizerTest.cpp.o
[ 99%] Building CXX object unittest/src/utils/common/CMakeFiles/testcommon.dir/FileHelpersTest.cpp.o
[ 99%] Building CXX object unittest/src/utils/common/CMakeFiles/testcommon.dir/StringUtilsTest.cpp.o
[ 99%] Building CXX object unittest/src/utils/common/CMakeFiles/testcommon.dir/RGBColorTest.cpp.o
[ 99%] Building CXX object unittest/src/utils/common/CMakeFiles/testcommon.dir/ValueTimeLineTest.cpp.o
[ 99%] Linking CXX executable ../../../../../../bin/testcommon
ld: unknown option: --start-group
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [../../bin/testcommon] Error 1
make[1]: *** [unittest/src/utils/common/CMakeFiles/testcommon.dir/all] Error 2
make: *** [all] Error 2
```

cmake output is:

```
-- Setting build type to 'Release' as none was specified.
-- The C compiler identification is AppleClang 10.0.0.10001044
-- The CXX compiler identification is AppleClang 10.0.0.10001044
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- CMAKE_BINARY_DIR: /Users/sumo/temp/sumo/build/cmake-build
-- CMAKE_SOURCE_DIR: /Users/sumo/temp/sumo
-- 
-- Platform: 
[...]

PR #4853, opened 2018-11-27
Fix macos cmake build; refs #4850

schwamborn:
Changes for cmake build to work under macOS


namdre:
@schwamborn  Please use the Signed-Off-By (git commit -s) mechanism to confirm the origin of the submission. The identity in the Signed-off-By must match the Author or the Committer.
    matthias.schwamborn@dlr.de

PR #5102, opened 2019-01-25
Merging Plexe platooning models into official SUMO release

michele-segata:
Dear all,

with this pull request, I would like to merge Plexe-provided platooning functionalities into the official SUMO release. I closed the previous PR because one commit was not signed-off.

Best,
Michele


behrisch:
Looks already great, but unfortunately it does not build under windows, mainly because you still use windows_config.h which we abolished recently. Please fix that. We also think it is probably a good idea to put the engine model in a different directory but since this is a little bit involved you can also leave this to us (although we are happy if you want to try :-)).

CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ endif ()
 # special debug flags
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -stdlib=libstdc++ -fsanitize=undefined,address,integer,unsigned-integer-overflow -fno-omit-frame-pointer -fsanitize-blacklist=${CMAKE_SOURCE_DIR}/build/clang_sanitize_blacklist.txt")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fsanitize=undefined,address,integer,unsigned-integer-overflow -fno-omit-frame-pointer -fsanitize-blacklist=${CMAKE_SOURCE_DIR}/build/clang_sanitize_blacklist.txt")
```
behrisch: What is the point in removing this? I think I remember that we had build problems with the clang libstdc especially on Mac.

behrisch:
> Looks already great, but unfortunately it does not build under windows, mainly because you still use windows_config.h which we abolished recently. Please fix that. We also think it is probably a good idea to put the engine model in a different directory but since this is a little bit involved you can also leave this to us (although we are happy if you want to try :-)).

I already fixed the windows_config.

CMakeLists.txt
michele-segata: Yes that is exactly the problem. On a Mac with clang, the -stdlib=libstdc++ causes the compilation to fail when including the <cstdlib> and the <random> libraries. This was the only way I've found to deal with this issue.

michele-segata:
Thanks for fixing the windows_config issue. I'm happy to move the engine models in a different folder. Suggestions for the name of the folder? I think it is better to have this settled to avoid future renaming.

CMakeLists.txt
behrisch: Are you bulding the "standard" brew way?

behrisch:
Simply src/microsim/engine

CMakeLists.txt
michele-segata: Actually I don't know what the standard brew way is :) I'm using `cmake` the following way:
```bash
mkdir build-release
cd build-release
CC=clang CXX=clang++ cmake -DCMAKE_BUILD_TYPE=Release ..
make
```

behrisch: I just meant the way you installed the dependent libraries because I think the main point is that all need to use the same cstdlib.

behrisch:
If you look here https://github.com/michele-segata/plexe-sumo/tree/master/src/microsim then there is no engine dir so I suppose you forgot to add the files.

michele-segata:
> If you look here https://github.com/michele-segata/plexe-sumo/tree/master/src/microsim then there is no engine dir so I suppose you forgot to add the files.

what an idiot... fixed. Hopefully everything is fine now.

behrisch:
It happens to the best of us ;-) Thanks for reacting so fast.

michele-segata:
Now it seems to have a problem with `cmake` on linux and windows machines. I tested the build on my mac, so I couldn't see this problem. I'll try to have a look today into this.

michele-segata:
Now the problem seems to be solved (for linux). However the TravisCI task failed not because of the code but it just hanged while running the `before_install` task. Is there a way (for you) to re-start the apple-autotools-gcc test?

michele-segata:
I have now fixed the windows compilation problems as well. The Travis build however failed for some reasons I don't understand. I don't think they are related to my code changes.
Do you know what the problem is?

behrisch:
We will work the rest out in the master. Thanks for fixing all the bits and pieces!

michele-segata:
> We will work the rest out in the master. Thanks for fixing all the bits and pieces!

Nice! Thank you for merging the code in.

PR #5107, opened 2019-01-27
Add access to PoI width, height, angle and background image in TraCI 

m-kro:
- PoI attributes width, height and angle are provided with getter/setter
- In addition to the mentioned attributes, a background image can be specified in traci.poi.add method
- Python and C++ interfaces are adapted to reflect these changes

Would it be possible to reserve a TraCI constant for the background image? Then I would add the corresponding methods, too...

### DIFF
--- a/src/libsumo/POI.cpp
+++ b/src/libsumo/POI.cpp
@@ -77,6 +77,24 @@ POI::getPosition(const std::string& poiID, const bool includeZ) {
 }
 
 
+double
+POI::getWidth(const std::string& poiID) {
+	return getPoI(poiID)->getWidth();
+}
+
+
+double
+POI::getHeight(const std::string& poiID) {
+	return getPoI(poiID)->getHeight();
+}
+
+
+double
+POI::getAngle(const std::string& poiID) {
+	return getPoI(poiID)->getShapeNaviDegree();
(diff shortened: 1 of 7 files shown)


namdre:
Thanks for the contribution! I've already added the new Constant.

 To improve downward compatibility with custom clients and (traci4matlab, veins, ...) and user code I would ask you for two amendments:
- reorder the parameters in the add() method so that all the new parameters come after the previous parameters. This way all the old code still works
- in TraCIServerAPI_POI.cpp please add a switch that accepts either 4 components or 8 and add the four new components to the end of the component list

Generally, we aren't too shy with braking the API but we still try to avoid it it isn't too much hassle.

Thanks,
Jakob

m-kro:
I was not able to test the changes completely yet as my SUMO compiled at work does not like jpg images?! But anyway, I hope the changes will be backward-compatible now.

m-kro:
Please check if all your requirements are met now. From my tests, both the old and new way for adding PoI work as does the getter/setter for the image.

m-kro:
Sorry for that. I have added the respective commits for the C++ client.

behrisch:
Great, thanks a lot!

PR #5147, opened 2019-02-04
TraCIConstants: use typed constexpr instead of #define

riebl:
Some generic names such as ADD and REMOVE easily clash when mixed with other source
code. Transforming old C-style #define constants to typed constexpr constants and moving
them to libsumo namespace eliminates this problem entirely.

Signed-off-by: Raphael Riebl <raphael.riebl@thi.de>


riebl:
Oh, no constexpr in VS2013, really?!

behrisch:
Thank you for the patch. I agree with the intention of removing the defines but this one does not work yet. @riebl Do you want to fix that or should we convert it into a ticket?

riebl:
Thanks for your feedback @behrisch! I have been unsure if my patch's intent is going to be welcomed by core devs at all. Next incremental patch is in the pipeline, stay tuned :-)

behrisch:
Thanks, this already looks good. There are some drawbacks however:
1. I would prefer to simply use int everywhere instead of the more specialized types. Any reason for introducing them?
2. We are trying not to use "using namespace". Can you remove those?
3. We need to adapt the tools/traci/rebuildConstants.py

src/traci-server/TraCIConstants.h
```diff
@@ -483,495 +498,498 @@
...
 // update bestLanes (set: vehicle)
-#define VAR_UPDATE_BESTLANES 0x6a
+TRACI_CONST ubyte_t VAR_UPDATE_BESTLANES = 0x6a;
 
 // filled? (get: polygons)
-#define VAR_FILL 0x55
+TRACI_CONST ubyte_t VAR_FILL = 0x55;
 
 // get/set image file (poi, poly, vehicle, person, simulation)
 #define VAR_IMAGFILE 0x93
```
behrisch: Why is this missing?

src/libsumo/libsumo.i
```diff
@@ -297,6 +297,7 @@ def simulationStep(step=0):
 %template(TraCIConnectionVector) std::vector<libsumo::TraCIConnection>;
 %template(TraCILogicVector) std::vector<libsumo::TraCILogic>;
 %template(TraCIStageVector) std::vector<libsumo::TraCIStage>;
+%include "traci-server/TraCIConstants.h"
```
behrisch: libsumo should not depend on traci-server but maybe we should move the constants to libsumo anyway?

src/traci-server/TraCIConstants.h
riebl: Oops, that's simply a mistake. I will fix this!

src/libsumo/libsumo.i
riebl: Moving the constants to libsumo sound reasonable to me. Shall I include it in this pull request?

riebl:
1) I thought it might be nice to have matching types w.r.t. TraCI protocol. I am fine with plain int as well, though.
2) Okay, have not been aware of that.
3) Ha, I have created something similar with [TraCIConstants.py](https://github.com/riebl/artery/blob/master/src/traci/sumo/traci-server/TraCIConstants.py) though for a different purpose. This script is not executed automatically by the build environment, is it?

src/libsumo/libsumo.i
behrisch: Yes.

behrisch:
1. Then please use plain int.
2. Yes.
3. No the script is only called on demand but it would be nice if it still works (for Python and Java).

riebl:
@behrisch Are you satisfied by the changes now?

behrisch:
Thanks a lot!

Issue #5182, opened 2019-02-12
Invalid `errno` assumption in StringUtils::toLong


mightyCelu:
The current implementation of `StringUtils::toLong` assumes that `errno` being `ERANGE` after calling `strtoll`/`strtoi64` means, indicates parsing an out-of-range literal and subsequently raises a `NumberFormatException`. However, the parsing functions set `errno` only when the literal actually is out of range. So when `errno` is already set to `ERANGE` when the function is called, the exception is thrown erroneously.

namdre:
fixed via  b2875d5c0dd48bf5e531e30f482e62e73a111be0

PR #5183, opened 2019-02-12
Reset errno before possibly writing to it

mightyCelu:
This fixes issue #5182 by explicitly setting `errno` to zero before calling the parsing functions.

This is the easiest workaround for the issue, an alternative, which might also reduce complexity, would be to replace the parsing functions with the STL functionality found in e.g. `stringstream`.


namdre:
Thanks for the patch! I'v merged it manually because of IP validation shenenigans. 
We're not using stringstream because it is much slower (by a factor of 40 according to http://www.kumobius.com/2013/08/c-string-to-int/)

PR #5348, opened 2019-03-21
Added color bar label option to plot_net_dump tool

tkchouaki:
Hi,
I thaught it would be useful to have the ability to specify to specify the label of the color bar produced by the **plot_net_dump.py** tool. So i modified the script to do so.

Now it is possible to launch it with the **--color-bar-label** option.


behrisch:
Thanks!

PR #5351, opened 2019-03-22
person generator tool

tkchouaki:
This tool allows to generate flows of persons for a SUMO simulation which is currently not possible in SUMO route files.
It does so by converting an xml file (usually having the ``.pflow.xml`` extension) to a sumo route file
containing the generated <peron> elements.
Here is an example ``.pflow.xml`` :
```xml
    <routes>
        <personRoute id="route-1">
            <walk from="e1" busStop="1" />
            <probability>
                <probabilityItem probability="0.5">
                    <ride busStop="2" modes="public" />
                    <probability>
                        <probabilityItem probability="0.5">
                            <stop busStop="2" duration="10" />
                        </probabilityItem>
                        <probabilityItem probability="0.5" />
                    </probability>
                </probabilityItem>
                <probabilityItem probability="0.5">
                    <ride busStop="3" modes="public">
                </probabilityItem>
            </probability>
        </personRoute>
        <personFlow id="forward" begin="0" end="3600" number="7" perWave="10" departPos="0" route="forward" />
        <personFlow id="backward" begin="0" end="3600" period="600" perWave="10" departPos="0">
            <walk from="e3" busStop="3" />
            <ride busStop="1" modes="public"/>
            <stop busStop="1" duration="50"/>
        </personFlow>
    </routes>
```
The example above allows to generate two flows of persons :

- The first flow consists of persons taking a bus from stop 1 to either stop 2 or stop 3 (with a 50% chance for each).
The persons of this flow are spawned in 7 waves (equally separated in time) and each wave consists of 10 persons.
For the persons going to bus stop 2, there's a 50% chance they'll stay there during 10 ticks.
The route followed by the persons of this flow is defined separately in a ``<personRoute>`` element and referenced by its ID.
- The second flow consists of persons taking a bus from stop 3 to stop 1 and then stopping there for 50 ticks.
The persons of this flow are spawned in periodic waves with 10 persons pere wave.
The route followed by the persons is defined directly under the ``<personFlow>``


behrisch:
Thanks a lot!

PR #5474, opened 2019-04-14
Drive on yellow signal: fine-grained configuration 

m-kro:
Currently, vehicles can only be configured to violate the **total** yellow signal duration by setting the junction model attribute jmDriveAfterRed to 0. Typically, the signal lasts between **3 and 5s** according to German guidelines. Given that even in the capacity formula of HBS 1s of yellow violation is "accorded", this would be a plus to configure that in SUMO. Additionally, treating the whole yellow time like green can lead to very short clearing times. So there is also a safety context...

This request adds an attribute **jmDriveAfterYellow** (>=0) which only is considered if **jmDriveAfterRed** attribute is deactivated (<0). The vehicle drives at yellow if the attribute value exceeds the current signal duration.


rlloretb:
How is SUMO supposed to behave by now with regard to yellow crossing?
I am finding no vehicle crosses on yellow (v.1.6.0)

namdre:
You should see vehicles crossing at yellow if they are already close to the stop line when the light switches from green to yellow.  The exact distance for this to happen depends on their speed.

rlloretb:
Weird. I set link speed to 50 km/h and I don't see any.
It also seems trajectories depart from red one second before green start.
v 1.6.0

namdre:
The visual timing of traffic light states lags 1 step behind the usual simulation state rules: Vehicles mostly use the state of the previous simulation step to plan their movements (what you saw in sumo-gui in the last step). However, the traffic light updates take place between the visualization of the last step and the movement update of the new step which means that the vehicles will react to the traffic light in the step on which you see it switch.

Nevertheless, if the vehicle is right in front of the stop line when the light turns yellow, then it will continue rather than stop.

There are other factors that can affect the likelihood of passing at yellow:
- step-length: by default, vehicles have constant speed during a simulation step. This gives them enormous braking capability because they can reduce their speed by 4.5m/s instantly and then travel with the reduce speed for the whole step. The effect is reduced by setting a lower step-length
- position update rule:  by setting option **--step-method.ballistic** the position update uses the average speed of the previous and current step. This reduces braking to realistic levels regardless of step length
- default deceleration. If you lower the value from it's default of 4.5 you will see more passing at yellow

rlloretb:
@namdre 
To compensate for the step mismatch between cars and light, I will add one extra second of orange phase.
I did decrease default deceleration and some trajectories now cross on yellow, but not many.
On the other hand, the position update rule worked perfectly.

Thanks for your reply!

PR #5493, opened 2019-04-19
Now it's possible to do : traci.simulation.getBusStopWaitingIDList

tkchouaki:
Hi,
I needed for my project to retreive the IDs of the persons at a bus stop via TraCI. Since it wasn't possible, i added that feature into  both SUMO and TraCi's source code.

- I added the Traci Constant : **VAR_BUS_STOP_WAITING_IDS** in both **SUMO**'s C++ code and **TraCI**'s python code (making sure to make no conflict)
- I added the method **MSStopingPlace::getTransportables** (to retrieve the persons)
- I added the method **Simulation::getBusStopWaitingIDList**
- I added a switch case to the **TraCIServerAPI_Simulation::processGet** method to handle the new **VAR_BUS_STOP_WAITING_IDS** constant.
- I finally added the **traci.simulation.getBusStoWaitingIDList** method in Traci's Python Code.

Regards,
Tarek CHOUAKI.

### DIFF
--- a/src/libsumo/Simulation.cpp
+++ b/src/libsumo/Simulation.cpp
@@ -270,6 +270,17 @@ Simulation::getBusStopWaiting(const std::string& id) {
     return s->getTransportableNumber();
 }
 
+std::vector<std::string>
+Simulation::getBusStopWaitingIDList(const std::string& id){
+    MSStoppingPlace* s = MSNet::getInstance()->getStoppingPlace(id, SUMO_TAG_BUS_STOP);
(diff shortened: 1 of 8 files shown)


behrisch:
Thank you for implementing this! Could you please change the code such that it does not create a list twice. Maybe already the stopping place could return a list of ids?

namdre:
I think a method that returns the transportables is cleaner in the long run and the extra list adds negligible overhead.

PR #5519, opened 2019-04-25
Now able to do : traci.vehicle[type].getPersonCapacity()

tkchouaki:
Hi,
I needed to retrieve the vehicle's capacity using TraCI. So i implemented this functionality.
In order to achieve that, i did the following:

- I added the TraCI constant : ** VAR_PERSON_CAPACITY = 0x38**. I checked, of course, that this value wasn't used before.
- I added the method getPersonCapacity() to both the classes : **libsumo::Vehicle** and **libsumo::VehicleType** and using this methos to hande the ** VAR_PERSON_CAPACITY** requests.
- I added the method **getPersonCapacity()** in both **vehicle** and **vehicleType** domains in TraCI python code.

Regards,
Tarek CHOUAKI.

### DIFF
--- a/src/libsumo/Simulation.cpp
+++ b/src/libsumo/Simulation.cpp
@@ -270,6 +270,17 @@ Simulation::getBusStopWaiting(const std::string& id) {
     return s->getTransportableNumber();
 }
 
+std::vector<std::string>
+Simulation::getBusStopWaitingIDList(const std::string& id){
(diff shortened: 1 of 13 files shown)


namdre:
Thanks for the contribution! I've added your patch manually because I didn't trust the squash-and-merge with all the extra commits in this pull request.

PR #5568, opened 2019-05-03
Vehicle::getNextTls() : fixing tls distance.

tkchouaki:
Hi,
I noticed two things when using :

```python
traci.vehicle.getNextTLS(vehicle_id)
```
- If the first next tls is on another edge than the vehicle, the returned distance corresponds to left distance to travel on the current edge.
- In every case, starting from the second TLS, they all have the same distance.

When i took a look into the code, i found that the **seen** variable was incremented only once (for the first edge).
I try here to solve this issue.


namdre:
Thanks for the patch! I had to tweak it just a little bit (adding seen after pushing the tls) but otherwise it was good.

PR #5620, opened 2019-05-10
Removing Transportable from Stop at waiting end

tkchouaki:
Hi,
i noticed that in cases like this:

```xml
<person id="person" depart="0">
        <walk from="1" busStop="busStop1" />
        <ride busStop="busStop3" modes="public" lines="line" />
        <stop busStop="busStop3" duration="30" />
    </person>
```
The person wasn't remove form the busStop at the end of the last stop element. & this would result in a SEGFAULT when trying to do traci.getBusStopWaitingIDList().

Here is a fix to this problem.

Regards.


namdre:
Thanks!

PR #5645, opened 2019-05-17
Minimum duration 0 in osmWebWizard Options panel

angelobanse:
Minimum value for the stepping interval of _"Duration"_ is now **0**.
If a negative number is inserted in _"Duration"_, the _type="number"_ invalidates the entry.

Signed-off-by: Angelo Banse <angelo_22@outlook.com>


behrisch:
@angelobanse Thank You!

PR #5654, opened 2019-05-19
Move debug code inside SSM device to avoid access to already deleted pointer

m-kro:
Encounter e can be deleted right before the current debug output code.


namdre:
Thanks for the fix, I added it manually because your commit was missing the signed-of footer.

PR #5671, opened 2019-05-24
Updates to tools/generateParkingAreaRerouters.py 

lcodeca:
- command line arguments added for automation purposes
- added default values for 
  - `--max-number-alternatives`, 
  - `--max-distance-alternatives`, 
  - `--min-capacity-visibility-true`, 
  - `--max-distance-visibility-true`


namdre:
cheers!

PR #5679, opened 2019-05-27
Added traci.vehicle.getActualBatteryCapacity()

tkchouaki:
Hello,

I added the possibility to retrieve the actual battery's capacity of vehicle using TraCi.

```python
traci._vehicle.getActualBatteryCapacity(vehicle_id)
```
Regards


namdre:
Thanks for the patch, however:
*  device parameter retrieval is already supported: https://sumo.dlr.de/wiki/TraCI/Vehicle_Value_Retrieval#Supported_Device_Parameters
* We don't want to add every device parameter as a separate traci call to avoid API bloat.

tkchouaki:
Thanks for the answer, i didn't notice that.

Issue #5495, opened 2019-04-21
add access to person stage information


behrisch:
much like route information for a vehicle the information of the different stages for a person plan should be accessible

namdre:
The existing person API dates from before the introduction of TraCIStage.
the current method person.getStage() should be changed from returning the stage type (int) to returning a TraCIStage object

tkchouaki:
I think there's a little problem,
In the TraCIStage class, the line attribute is a single string. while a riding stage can have multiple desired lines.

namdre:
Right now, sumo parses the list of lines from a string (when reading xml) with the hard assumption that SPACE is the separating character. Therefore, multiple lines can also be returned as a single string by joining them with SPACE.

PR #5695, opened 2019-05-29
traci.person.getStage() now returns a Stage object unstead of a number

tkchouaki:

Hi,
This should sove #5495 
but for now, i am just setting the cost, the length and the travelTime of the stage to -1.

Regards, 
Tarek CHOUAKI.


namdre:
Thanks!

PR #5710, opened 2019-06-04
edit to add until parameter into setStop function in TraaS

YenKang:
Add edited codes in **Vehicle.java** and **Traci.java** to finish **"until"** function in **setStop**

```java
String edge1 = "41389174#2";
double pos1 = 10.0;
double until_1 = 500.0;
SumoStopFlags sf1 = new SumoStopFlags(false, false, false, false, false);
conn.do_job_set(Vehicle.setStop("1", edge1 , pos1, (byte)0,  0.0,  sf1, pos1, until_1));
}
```

100s-stop


behrisch:
Thanks for the pull request but shouldn't the code modify setStop instead of setParkingAreaStop?

YenKang:
I tried to make it work in setStop function first, and **setParkingAreaStop** maybe have the similar logic.

PR #5717, opened 2019-06-05
Ability to retrieve chargingStations's lane id.

tkchouaki:

Hi, added the ability to retrieve a chargin station's lane id via traci.

```python
traci.simulation.getParameter(station_id, "chargingStation.lane")
```
Regards.


tkchouaki:
I noticed that the ChargingStation class inherits from the Parametrised class but it is not really used. Is there a reason for that ? is it planned to change the code so it will be used more ?

namdre:
Generic parameters are mainly used to support user specific extensions.
They can be shown in the GUI and also retrieved via TraCI now (#5720, #5721)

Issue #5498, opened 2019-04-22
allow adding TraCIStage directly to a person plan


namdre:
The existing person API dates from before the introduction of TraCIStage.
person.appendStage(personID, traciStage)

possibly makes #4156 obsolete

PR #5768, opened 2019-06-25
traci._person.appendStage

tkchouaki:
This is an attempt to solve #5498 while maintaining compatibility for old TraCI requests.


namdre:
Thanks!

PR #5764, opened 2019-06-25
Updated README

sunn-e:
added direct download link.
Formatted the document according to standard.

PR #5771, opened 2019-06-26
Documentation fixes and added direct download links. 

sunn-e:
Thanks for the guidance about ECA @namdre  I hope I am doing this correct.
You may refer to #5764 for further comments.

Signed-off-by: Sunny Dhoke <sunnydhoke22@gmail.com>


sunn-e:
Still ECA failed? What am I doing wrong?

namdre:
The 'updating readme' commit seems to be fine.
However, the other two merge commits are failing validation. This time Eclipse doesn't say why and just splits out `This validation request is closed or unavailable.`
I'm just guessing at this point but it might help to redo the contribution as a single commit from a freshly forked repo.

behrisch:
I like the repo size badge but I don't like the reformatting. What is the reason to do so? Especially adding the backtick formatting for the command seems to increase indentation and makes it harder to read. Also I prefer the visual difference between dashes and equal signs for headlines which is better to read than counting the number of hashes.

sunn-e:
I just followed the markdown standards for GitHub. The number of hash defines the level of header. 1 hash defines it is highest level title and consecutively adjusting hash helps to keep track of level.
 # Biggest 
## Smaller
 ### Getting smaller
#### Much smaller 
##### and like wise

sunn-e:
I was suggested to follow the same by OWASP mentors.

behrisch:
Thank you for the request. I just included the badge and fixed some links. Our primary download location will change soon, so I did not include that but will keep it in mind. If you have more ideas for improvements please open a new request.

sunn-e:
No problem @behrisch , thanks for informing. I'll send a new PR if something new comes to my mind.

PR #5774, opened 2019-06-27
traci.person.setParameter("personId.stageIndex", "drivingStage.Lines"…

tkchouaki:
Hi,
I added the possibility to change the **lines** attribute of a driving stage online by using :

```python
traci.person.setParameter("personId.stageIndex", "drivingStage.Lines", "lines")
```

The given id should be composed of the id of the person & the concerned stage index.

Regards.

### DIFF
--- a/src/libsumo/Person.cpp
+++ b/src/libsumo/Person.cpp
@@ -28,6 +28,7 @@
 #include <microsim/pedestrians/MSPerson.h>
 #include <libsumo/TraCIConstants.h>
 #include <utils/geom/GeomHelper.h>
+#include <utils/common/StringUtils.h>
 #include <utils/common/StringTokenizer.h>
 #include <utils/common/SUMOTime.h>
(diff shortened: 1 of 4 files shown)


behrisch:
Thank you for the request but I think I would rather not follow the path of adding a separate call for every parameter. Wouldn't it be better to let people retrieve the stage, modify it and send it back?

tkchouaki:
The current API only allows appending a stage at the end of a person's plan.
Should the process be like this: 

- Retrieve the stage
- Remove it from the person's plan
- Modify the desired parameters
- Append the stage (so the APPEND_STAGE request should take also a stageIndex parameter)

Or maybe:

- Retrieve the stage
- Modify the desired parameters
- Replace it in the person's plan (by adding a REPLACE_STAGE request)

behrisch:
I would prefer the second alternative.

namdre:
I'm also in favor of adding a REPLACE_STAGE request

PR #5790, opened 2019-07-02
get rid of the hack in MSLane to inform MSCFModel_CC when a collision occurred

michele-segata:
This commit removes the need of referring to the specific MSCFModel_CC class within the MSLane class to inform the car-following model when a vehicle has collided. The MSCFModel_CC class now automatically queries for stops and checks whether a stop has been caused by a collision.
This change properly works when the parameter `collision.action` is set to `remove` and the parameter `collision.stop` is set to a certain number of seconds.

Signed-off-by: Michele Segata <segata@ccs-labs.org>


behrisch:
Thanks!

PR #5833, opened 2019-07-17
Initial updates to introduce manoeuvering behaviour into vehicle parking:

theDivj:
Uses the angle between the lane and parking bay/lot as a proxy for the complexity of the parking manoeuvre.
For off-road lots the angle is computed during lot entry - ie the complexity of the parking space is a 'lot' attribute.

The time taken to manoeuvre through given angles is a map held as a configurable vtype attribute.
( A bus/coach/truck will take longer to park than a passenger car) using the lot angle and the map.

The premise was to keep the vehicle in the lane whilst entering/exiting the lot in order to assure that 
representative lane blocking behaviour and associated emissions etc could be captured.

Update includes the current 'sumo/extended/rerouter/parking' tests reflected into a subfolder with the manoeuvre modelling turned on - 
this shows the impact of the behaviour on each test - in all cases behaviour differences are associated with
additional waiting introduced by the parking timing.

Attached illustrates the kind of parking area I'm trying to model - very promising to date.
detailSnap


theDivj:
closing to recreate with signed off tags

PR #5841, opened 2019-07-22
Cache for rerouters generator

lcodeca:
Added:
- info on the status
- progress bar during the generation
- cache for getShortestPath to avoid multiple computations for the same edges

Imports cleanup.


behrisch:
I very much like the idea of doing this with cache and without traci but could you please make tqdm optional?

lcodeca:
Sure, now tqdm is optional.

behrisch:
Thank you!

Issue #5527, opened 2019-04-26
SSM device: questionable TTC and PET values


m-kro:
I evaluate conflict situations at a signalized intersection using the SSM device. Recently I have come across some strange values for TTC and PET measures I don't think they are intended:

- min TTC  of 0 is calculated for some vehicles when they approach the queue in front of the traffic light. However from looking at the GUI they do not collide. But minGap violation is not considered here, right?
- very negative PET values (probably "-INVALID"?) are printed to the output value

Both phenomenons can be observed in the same example simulation attached...
[results_K048_Festzeit_Nullfall1_ssm.zip](https://github.com/eclipse/sumo/files/3122308/results_K048_Festzeit_Nullfall1_ssm.zip)

namdre:
Thanks for the input example.  From the description alone it looks like a duplicate of #4569 and #5033.

m-kro:
Some indication from the "ssm.trajectories" output: TTC=0 happens when both vehicles have come to a standstill (velocity vector -0.000,-0.000). For most time steps, "NA" is returned, but 0 for others. All in all looks like a numerical problem.

m-kro:
Update: computeTTC is called with negative gap values which trigers ttc=0.

m-kro:
We're getting closer: It seems SSM device does not consider split connections properly when calculating the distance. In my example, the gap between two vehicles drops sharply when the first touches the ped crossing. The variable egoDistToConflictLane stays at 0 for an ego vehicle waiting at the traffic light upstream.

m-kro:
Snippet of the SSM DEBUG output showing the last timestep with NA value and the first with TTC=0. Veh. Pkw1.4 wants to follow Pkw1.2 and turn right but stops at the red signal. Pkw1.2 reaches the stopping position at the pedestrian crossing.
`
135.80 updateEncounter() of vehicles 'Pkw.1.4' and 'Pkw.1.2'
classifyEncounter() called.
135.80 findFoeConflictLane() for foe 'Pkw.1.2' on lane ':K048_Nord_0_0' (with eg
oConflictLane=:K048_Nord_0_0)
foeBestLanes: NULL gneE54_0 -gneE7_0
Potential conflict on junction 'K048_Nord
egoConflictLane: ':K048_Nord_0_0'
foeConflictLane: ':K048_Nord_0_0'
Ego's distance to conflict lane: 0.00
Foe's distance to conflict lane: -20.59
[...]

PR #5845, opened 2019-07-23
Float zero comparison in CFM arrival estimation

m-kro:
Replaces the direct floating point comparison to 0 inside the CFM arrival estimation with a check against NUMERICAL_EPS.

### DIFF
--- a/src/microsim/cfmodels/MSCFModel.cpp
+++ b/src/microsim/cfmodels/MSCFModel.cpp
@@ -390,7 +390,7 @@ MSCFModel::estimateArrivalTime(double dist, double speed, double maxSpeed, doubl
     assert(speed >= 0.);
     assert(dist >= 0.);
 
-    if (dist == 0.) {
+    if (dist < NUMERICAL_EPS) {
         return 0.;
     }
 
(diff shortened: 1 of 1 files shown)


behrisch:
@m-kro Do you have a small example where it makes a difference?

m-kro:
This is connected to the ongoing corrections in the SSM device, where the CFM arrival estimation is used. I actually had trouble with a very very small acceleration value. I will need some time to compile a working example.

behrisch:
Thanks, it wouold be great to have one.

m-kro:
Ok I have found the simulation where I discovered this. It is the one from issue #5527 using seed 27. At time 2402.2, the [assertion eInfo.foeEstimatedConflictEntryTime > 0.](https://github.com/eclipse/sumo/blob/336991261408185c2e6115eb9687c7bf89df02f5/src/microsim/devices/MSDevice_SSM.cpp#L902) is triggered for the conflict ego "Pkw.28.8" vs. foe "Pkw.32.45".  eInfo.foeEstimatedConflictEntryTime is the outcome of the CFM arrival time estimation.

namdre:
I was unable to reproduce the assertion. However, the change looks reasonable enough and does not break any tests so I'm merging it anyway.

PR #5847, opened 2019-07-24
TraCIAPI: remove inclusion of config.h header

riebl:
No defines from config.h are actually used by TraCIAPI at all.
Furthermore, TraCIAPI should be independent of a particular SUMO build just as TraCIDefs.h.

Signed-off-by: Raphael Riebl <raphael.riebl@thi.de>


behrisch:
Thanks!

PR #5857, opened 2019-07-27
Revert c1e6f65f2b28816e52eefc05fb7e5f176c703b26 on this file only to …

theDivj:
…fix duarouter etc.

The memory leak changes fixes ref#12 broke duarouter, dfrouter etc - replacing this line restores the status quo.  (there were significant failures in the texttest suite)

(I don't understand the code well enough to understand why this fixes the issue - so alternatives may well be feasible.)


theDivj:
Additional commits introduce modeling of parking manoeuvres and associated tests.

Uses the angle between the lane and parking bay/lot as a proxy for the complexity of the parking manoeuvre. For off-road lots the angle is computed during lot entry - ie the complexity of the parking space is a 'lot' attribute.

The time taken to manoeuvre through given angles is a map held as a configurable vtype attribute.
( A bus/coach/truck will take longer to park than a passenger car) using the lot angle and the map.

The premise was to keep the vehicle in the lane whilst entering/exiting the lot in order to assure that
representative lane blocking behaviour and associated emissions etc could be captured.

Testing commit contains the 'sumo/extended/rerouter/parking' tests reflected into a subfolder with the manoeuvre modelling turned on - this shows the impact of the behaviour on each test - in all cases behaviour differences are associated with additional waiting introduced by the parking timing and in one case where the order of vehicles changes because of the entry/exit delays.

theDivj:
closed to resolve conflicts

PR #5873, opened 2019-08-05
remove pointer usage in SUMOSAXReader, fix #5871

### LINKED ISSUE #5871: SUMOSAXReader is leaking file handle
I noticed that Sumo is leaking file handles on the route file definition when using it with libsumo. 
Running a simulation with libsumo in a loop and starting/closing libsumo via traci will generate file handles on the route file in read mode, which will never get released.

This happens with the latest version from this repository, the latest release 1.2 works fine. Happened for me on Ubuntu 18.04. 

It looks like the bug was introduced by this [commit](https://github.com/eclipse/sumo/commit/5dc5f9fa1ef66207ec9b151963842b988fa7a696)

The problematic code is this one:
```
#ifdef HAVE_ZLIB
        myIStream = new zstr::ifstream(systemID.c_str(), std::fstream::in | std::fstream::binary);
        myInputStream = new IStreamInputSource(*myIStream);
        return myXMLReader->parseFirst(*myInputStream, myToken);
#else
```

You are allocating your file input streams with _new_ and then leak the pointers, the xmlReader is not owning those and will not clean up. As a result the file referenced in systemID will never be closed again.

I think doing something simple like this should fix the issue:
```
bool parsed = myXMLReader->parseFirst(*myInputStream, myToken);
delete myInputStream;
delete myIStream;
return parsed;
```

Is there any reason why you need the pointers there in the first place though?
I guess you accept PRs? I can open one then with something along the lines of the change suggested above.

For reproducing the issue I adapted your example from here: tests/complex/traci/clientClose as follows:
```

from __future__ import absolute_import
from __future__ import print_function

import os
import subprocess
import sys

import psutil

if 'SUMO_HOME' in os.environ:
    tools = os.path.join(os.environ['SUMO_HOME'], 'tools')
    sys.path.append(tools)
else:
    sys.exit("please declare environment variable 'SUMO_HOME'")

import libsumo as traci  # noqa



def runSingle(traciEndTime):
    step = 0
    traci.start(["-c", "sumo.sumocfg"])
    while not step > traciEndTime:
        traci.simulationStep()
        step += 1
    proc = psutil.Process()
    print("# file handles open", len(proc.open_files()))
    #for file in proc.open_files():
    #    print(file.path)
    traci.close()
    sys.stdout.flush()


print("=========== long route ===========")
fdo = open("input_routes.rou.xml", "w")
print('<routes>"', file=fdo)
print(
    '   <route id="horizontal" edges="2fi 2si 1o 1fi 1si 3o 3fi 3si 4o 4fi 4si"/>', file=fdo)
print('   <vehicle id="horiz" route="horizontal" depart="0"/>', file=fdo)
print('</routes>', file=fdo)
fdo.close()

print("----------- Start traci -----------")
sys.stdout.flush()
for i in range(10000):
    runSingle(10)
print("----------- done -----------")
```

If everything works correctly you should see this output:

> # file handles open 0

Instead with the current version from git, you will see the open file handle count being incremented every run

jacobowitz:
This PR should fix the file descriptor leak described [here](https://github.com/eclipse/sumo/issues/5871).

I just replaced the usage of the raw pointers in the parseFirst method and use shared pointer instead, which will clean up the file handle by default.

I don't see a reason for using raw pointers here.
I also removed them as members of the class as they are not needed as well. The fields are never used and can't be used for later clean up in the destructor as parseFirst is public and could be called multiple times.

### DIFF
--- a/src/utils/xml/SUMOSAXReader.cpp
+++ b/src/utils/xml/SUMOSAXReader.cpp
@@ -24,6 +24,7 @@
 #include <config.h>
 
 #include <string>
+#include <memory>
 #include <iostream>
 #include <xercesc/sax2/XMLReaderFactory.hpp>
 #include <xercesc/framework/LocalFileInputSource.hpp>
@@ -158,8 +159,8 @@ SUMOSAXReader::parseFirst(std::string systemID) {
         }
         myToken = XERCES_CPP_NAMESPACE::XMLPScanToken();
 #ifdef HAVE_ZLIB
-        myIStream = new zstr::ifstream(systemID.c_str(), std::fstream::in | std::fstream::binary);
-        myInputStream = new IStreamInputSource(*myIStream);
+        myIStream = std::unique_ptr<zstr::ifstream>(new zstr::ifstream(systemID.c_str(), std::fstream::in | std::fstream::binary));
+        myInputStream = std::unique_ptr<IStreamInputSource>(new IStreamInputSource(*myIStream));
         return myXMLReader->parseFirst(*myInputStream, myToken);
 #else
         return myXMLReader->parseFirst(systemID.c_str(), myToken);
--- a/src/utils/xml/SUMOSAXReader.h
+++ b/src/utils/xml/SUMOSAXReader.h
@@ -26,6 +26,7 @@
 #include <config.h>
 
 #include <string>
+#include <memory>
 #include <vector>
 #include <xercesc/sax2/SAX2XMLReader.hpp>
 #include <xercesc/sax/EntityResolver.hpp>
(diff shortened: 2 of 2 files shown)


behrisch:
Also please open an Eclipse Account (if you do not already have one) and sign your commits as mentioned here: https://sumo.dlr.de/wiki/FAQ#How_do_code_contributions_work.3F
Sorry for the hassle.

jacobowitz:
@behrisch 
Yes the InputStream will be invalidated after the function returns. So I figure that this is a problem for parseNext() etc, right?
Will change the PR then to respect that fact and will create the Eclipse account

jacobowitz:
Updated this pr:
- This PR now just wraps the previous raw pointers in unique pointers
- Registered with eclipse and signed off the commit

@behrisch 
I registered and signed off on this commit, but it kinda looks like it is not working from the link? Maybe force pushing over my existing commit didnt work?

behrisch:
Thanks, looks much better now. Did you already sign the ECA after creating the Eclipse account, because that is what the check is complaining about?

jacobowitz:
@behrisch 
looks like I broke the signoff check somehow for this pr (I guess the eclipse integration got confused with the force push)
Let's use [this pr](https://github.com/eclipse/sumo/pull/5911) instead
Seems to be working in the new pr

Issue #5644, opened 2019-05-17
False surrounding vehicles scan order for SSM device


leoluecken:
Scanning recursively for all upstream vehicles may confuse some downstream conflicts since the ego conflict lane is not recognized correctly (but set to its current lane). Downstream scan along the ego's route should complete before executing upstream scans.

m-kro:
In turn, this solution may as well cause wrong ego conflict lanes itself due to the scan order, see #5527

m-kro:
Currently, all internal lanes of a junction get assigned the same ego conflict lane, even if they are not foes of the ego route. If such a lane leads to a separate crossing with the ego route, still the internal lane of the other intersection is chosen.

behrisch:
@m-kro Do we need another ticket for that (or reopen this one?)

m-kro:
I would reopen the ticket as its title is still in line with the new findings.

PR #5877, opened 2019-08-06
SSM foe search and scan order

m-kro:
As mentioned in issue #5644  and #5527 , the current scan for foe candidates assigns the wrong conflict lane in some cases. This results in utterly underestimated TTC values. For more details look at the referenced issues. 
This pull request changes the recursive search to stop at visited (internal) lanes instead of visited junctions. Additionnally, inside junctions only foe lanes of the ego route are addressed. This way, the right conflict lane can be propagated through the network. The redesigned search mechanism still lacks some proper handling of pedestrian crossings which makes it less universal than the current one.
Please review the pull request thouroughly and give me hints in which sense it should be updated.

### DIFF
--- a/src/microsim/devices/MSDevice_SSM.cpp
+++ b/src/microsim/devices/MSDevice_SSM.cpp
@@ -3003,7 +3003,7 @@ MSDevice_SSM::getUpstreamVehicles(const UpstreamScanStartInfo& scanStart, FoeInf
         lane->releaseVehicles();
 
 #ifdef DEBUG_SSM_SURROUNDING
-		if (gDebugFlag3 &&& foundCount > 0) {
+		if (gDebugFlag3 && foundCount > 0) {
 			std::cout << "\t" << lane->getID() << ": Found " << foundCount << "\n";
 		}
 #endif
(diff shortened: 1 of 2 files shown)


namdre:
Thanks for working on this! It really helps when I don't have to do this alone (-:
It will take me a while to review this patch since Its been a few weeks since I looked at the SSM-code in detail. You may want to run the sumo/tests/sumo/devices/ssm suite yourself.

In it's current state,  the patch breaks 25 of of our 89 ssm-device tests.
* slight variation in timing (which might be an approvement or not)
* N/A values for position/conflict  type/value (probably a bad sign)
* one test crashes

m-kro:
Could you post the name of the test which crashes? With my local repository, I get 24/88 failed tests.

namdre:
The crashing test is devices/ssm/bugs/ticket5839 (this test was only added 8 days ago).

namdre:
Running astyle as part of our release cycle broke the patch, sorry.
Could you please fix the pull request on your end?
Did you get the change to look at the failing tests and check whether the new N/A values are appropriate?

m-kro:
So far I can only confirm there is a problem with my code with respect to overtaking on the opposite lane. I will work on it during the next days and get back to you.

namdre:
There is still a merge conflict. I think the easiest way to resolve this is to make a new pull request that replaces our astyled version with your previously patched version (thereby reverting our astyle changes).

m-kro:
OK so I close this pull request and go back to work. I have solved most of the failing tests by now.

PR #5894, opened 2019-08-08
Add optional parking entry/exit times set by vType parameter

theDivj:

Code to optionally allow more realistic parking space entry/exit timings (as-is enters/leaves space in a single sim step) meaning lanes are blocked for longer than a single step for parking, obscuring parking related congestion.

Timings are set by a vType parameter of the form:
manoeuverAngleTimes="10 3000 4000,80 1000 11000,110 11000 2000,170 8000 3000,181 3000 4000"

Each triplet in the parameter consists of an angle value followed by an entry time and exit time.  The angles delimit ranges corresponding to the angle between the space and the lane.

In the example above a space that is angled between 0 and 10 degrees from the lane will block the lane for 3 seconds before entering the parking space and for 4 seconds on leaving the space.


namdre:
Some comments and questions:
- Thanks for the contribution!
- Big bonus points for adding test cases to your pull request!
  - it would be good to include the new vType parameter in your test

- the compilers throws plenty of warnings "type qualifier ignored on function return type" whenever you return things by value
- xml input parametes are all in seconds rather than milliseconds by convention

- I'm not convinced it is usefull to model parking maneuvers for parkingAreas with individual `<space>`-definitions. In my mind such a parkingArea just models the entry to a car park and we ignore all the internal dynamics (extra roads etc.). Thus the angle is not meaningful since it does not relate to the lane of the parkingArea but rather to roads which aren't part of the simulation. In contrasts, parkingAreas with roadsideCapacity model parking next to the road at a configurable angle. 
  - I realize you've used this to model forward and reverse parking at the same roadside parkingArea in your test but I would argue that to the driver, all lots look the same and it's a driver decision whether to use forward or reverse parking rather than a feature of the lot. (And thus the same situation should be modelled with two different driver types rather than two different space angles)
  - all the code in addLotEntry could be simplified.
  - no need for myManeuverAngle as MSParkingArea::myAngle captures the relevant information

- Please explain why myEndPositions can grow beyond capacity. What does that mean?
  - Also, please explain in which situation a swap occurs
  
- in planMoveInternal, I would rather apply the exitManeuver-speed-restriction to `double v` rather than `const double maxV`

- I don't understand why the extra check in the lane change models is needed. The returned value is MSVehicle::myStopDist which is always defined even when the vehicle has no stops. Also, since your patch does not affect myStopDist everything should work as before (At least I need another test case to convince me this change is necessary)

theDivj:
Many thanks for looking at this - responses inline:-

>it would be good to include the new vType parameter in your test

doh - of course - I'd cut this out when i was shrinking the submission :(

> the compilers throws plenty of warnings "type qualifier ignored on function return type" whenever you return things by value

I did say my C++ was rusty :) - I've no excuse for not checking warnings though - will rework!

> xml input parametes are all in seconds rather than milliseconds by convention

they are ms (SUMOTime) - only my explanatory text in the pull request used seconds

----
> I'm not convinced it is usefull to model parking maneuvers for parkingAreas with individual <space>-definitions.

The code works in the same way for laneside parking areas. My specific interest was modelling behaviour in a school parking lot - which I am doing by having many lots alongside lanes. I've been pleasantly surprised by the modelled behaviour already approximating to observed - though i have lots of tuning and proving to do before I can use any results.
In real-life the lot I am modelling has significant impact on the surrounding roads as traffic builds up at the entry, with something close to a deadly embrace happening when the exit (500m away) is blocked by entering traffic - and teleport isn't a real-life option. :) I suspect that modelling this kind of congestion hit ought to be more generally useful.

----
>I realize you've used this to model forward and reverse parking at the same roadside parkingArea in your test but I would argue that to the driver, all lots look the same

The trigger for me to look at this was a change in a school parking area space layout where the one-way system and bay orientation was forcing 'nose-in' parking. I feel this is not appropriate for a school car park as this means that the peak of reversing is when most children are traversing the car park. Subjectively this also slows departure as vehicles are all trying to depart within a relatively short time interval.

Yes the choice of reverse/forward in a 90 degree bay is down to driver preference/park rules. However a bay at an acute angle to the lane (the standard 45 or 70 degree bays) promotes nose-in parking, it is quite difficult to reverse-in to such a bay. Equally, leaving such a bay after reversing in involves a much wider turn.  The opposite considerations apply to an obtuse angled bay.

ie Driver type is only relevant to bays at roughly 90 degrees to the lane.
( In my modelling I am currently assuming that everyone behaves the same for 90 degree bays as the reversing manoeuver has to happen at either entry or exit. It does not feel worthwhile to model eg slower reversing when adjacent bays are full versus empty. It's probably more important, for bays parallel to the lane, to model parallel parking vs straight in, when occupancy permits.)

There are several UK council guidelines that recommend bay orientation to promote reverse parking, a number of safety conscious large corporates mandate this and there have been several US trials but I cannot find anything that is evidential or citable - hence my wish to simulate this and determine whether I can illustrate differences.

----
> Thus the angle is not meaningful since it does not relate to the lane of the parkingArea but rather to roads to roads which aren't part of the simulation. In contrasts, parkingAreas with roadsideCapacity model parking next to the road at a configurable angle.

  I do not disagree - I'm only using spaces to get a closer visual fit to the parking area. (cf attached .png) 
  Where the parking area lanes are straight, then using a laneside area is fine, the manoeuvering angle is then the area angle as you say.

----
> all the code in addLotEntry could be simplified
> no need for myManeuverAngle as MSParkingArea::myAngle captures the relevant information

Agreed, but that precludes the model being applied to <space> lots as the space angle is absolute rather than relative to the lane?  I'd like to model the behaviour 'inside' the parking lot - microsimulation feels appropriate - I'm not sure there are any alternatives out there - but I'm a novice in this domain.

----
> Please explain why myEndPositions can grow beyond capacity. What does that mean?
> Also, please explain in which situation a swap occurs

Attached zip demonstrates these - please see the readme.txt

----
> in planMoveInternal, I would rather apply the exitManeuver-speed-restriction to double v rather than const double maxV

understood - I think I looked at this but it involved applying changes in several places - will look again!

----
> I don't understand why the extra check in the lane change models is needed. The returned value is MSVehicle::myStopDist which is
always defined even when the vehicle has no stops. Also, since your patch does not affect myStopDist everything should work as before (At least I need another test case to convince me this change is necessary)

I was getting exceptions under a busy parking scenario - I think it was before I'd recognised the need for the endPosition fix - i'll revert code and see if I can get a test case together.
[endpos.zip](https://github.com/eclipse/sumo/files/3486826/endpos.zip)

----
In the meantime I'll close the pull and rework.

cheers and thanks again
div
detailSnap

namdre:
> > xml input parametes are all in seconds rather than milliseconds by convention
> 
> they are ms (SUMOTime) - only my explanatory text in the pull request used seconds
> 
My complaint was ambiguously worded, it seems. What I meant to say is: The conventional unit for all xml time inputs in SUMO is seconds. Please change the milliseconds to seconds.

> I do not disagree - I'm only using spaces to get a closer visual fit to the parking area. (cf attached .png)
> Where the parking area lanes are straight, then using a laneside area is fine, the manoeuvering angle is then the area angle as you say.

There are more differences between using roadsideCapacity vs `<space>`.
For the former, the position along the lane that vehicles drive to for parking corresponds to the visual position of the lot. When using `<space>`, vehicles always drive to the endPos of the parkingLot (because this is intended to model the entrance position).

Thus, using `<space>` just to fine-tune appearances isn't really supported.
A possible solution would be to define multiple `<parkingArea>`s instead of one to have more control over angle and position (with attendand parkingAreaRerouters).

> > Please explain why myEndPositions can grow beyond capacity. What does that mean?
> > Also, please explain in which situation a swap occurs
> 
> Attached zip demonstrates these - please see the readme.txt
> 
Thank you for the example. The behavior was due to #5905. The test also exposed #5907. Both issues are now fixed.

theDivj:
Thanks again, I appreciate your guidance - I quite understand that parking is not overly significant in the overall modelling context!  ( Seconds/ms understood and updated)

I originally set my model up using discrete parking areas because I wanted to incorporate the preference for places nearest the school paths or nearest the exit. This didn't work but mainly because of issue 5864. 
I have now rearranged the model to use parking areas but it does produce anomalous behaviour:
Specifically:  
a) Parking vehicles travel to the end of the area before 'jumping back' into a space.
b) Vehicles wanting to park in a currently 'full' area travel to the end of the area, where you have the strange behaviour of an exiting vehicle from earlier in the area jumping in front and the waiting vehicle then jumping back into the area.

I can minimise this by using multiple small areas however the original (pre-fix) behaviour, where the vehicle inserted at the position it left the lane, was more natural.  More problematic is that I'm not getting the congestion behaviour I had before, which actually reflected my observations of the real park. I'll play around a little more before reverting to <space> though.

In this context - is there a specific rationale/model limitation that means the end of the area needs to be used as the targetPos? In reality vehicles either reroute or wait at the beginning of an area. Stopping at the beginning would at least allow the previous insertion behaviour to work without causing a jam?

cheers
div

namdre:
According to the original implementation idea of parkingAreas, the `<space>` elements have no geometrical relationshop to the parkingArea. An example would be multiple rows of space in a car park next to road. The parkingArea object is meant to be short (e.g. 3m) and model the entrance to a car park. In this case jumping back is not an issue. The use case for this setup is where you don't want to model the dynamics within the car park (lots of tiny roads) and still want to see at a glance how many vehicles are parked there.
If you don't care about the visuals, you can just defined a 3m long parkingArea with roadsideCapacity=100 and it will work.

If you need a stretch of space along a road where the vehicles enter/exit according to the position of the rendered lot, you can define this with a long parkingArea and roadsideCapacity. In particular, issues a) and b) will not occur when modelling parking in this way.
If the lots don't look quite the way you want because the road is curved or there are some lots with different angle, you can just define multiple parkingArea elements along the same lane (again with roadsideCapacity).  

The target postion for leaving and entering is lsd.myEndPos (and the waiting position for other vehicles is computed relative to that). In master, lsd.myEndPos is always set to parkingArea::myEndPos when using `<space>` (for the reason outline above). 
However, your patch (as taken from your prior pull request) maps this to the closest position along the lane instead so it should actually work just like roadsideCapacity.
Since it doesn't hurt the original useCase that's just great. (And I wouldn't even make it conditional on gModelParkingManeuver)

Your comment sounds as if this doesn't work, though. If my comments are insufficent help to figure out why, feel free to send another scenario (and code patch).

theDivj:
The current build - presumably with the #5905/7 fixes works very well indeed - thank you.
I understood your description of the behaviour of non <space> areas but I was getting odd blocking/jam behaviour with <space> areas when I applied my changes to this build.

The cause was the fact that the reservation code has an implicit dependance on the first <space> being nearest the beginning of the parking area. Once I reflected that, everything works very well indeed - and I can still use the positions of other spaces to reflect area preferences and no longer need teleport to complete a model run. Once again many thanks.

I think I've now caught all your comments in this codeset - obviously feel free to comment, tear apart, ignore etc!

cheers

PR #5900, opened 2019-08-09
Adjustments for external inclusion

DannyBehnecke:
These changes are for including SUMO into another CMake based project. The attached CMakeLists adds some glue lines for everything to work accordingly and to be put in a separate folder in visual studio. 
[CMakeLists.txt](https://github.com/eclipse/sumo/files/3486718/CMakeLists.txt)

I mostly changed CMake variables to the LIST variant. Also, when you set the GENERATED property for a header file, only the listed argument gets assigned this property, not the file the argument is pointing to. Example: 
a) set_source_files_properties(typemap.h PROPERTIES GENERATED TRUE)
b) set_source_files_properties(foo/typemap.h PROPERTIES GENERATED TRUE)
is not the same. This can be checked by querying the properties of typemap.h . In case a) it will be set to TRUE in case b) it's just an empty string.


behrisch:
I included the things which looked like they could not break anything but left out the parts which looked suspicious. Could you recheck whether it already suffices and if not do another pull request?

behrisch:
About the generated property: The cmake docs state that output files of custom targets get this property automatically.

PR #5911, opened 2019-08-12
replace raw pointer with unique_ptr in SUMOSAXReader

jacobowitz:
new PR to replace the one with the broken sign-off

fixe file handle leak in SUMOSAXReader by replacing raw pointers with unique pointers.

Signed-off-by: Tobias Jacobowitz <tobias.jacobowitz@posteo.de>

### DIFF
--- a/src/utils/xml/SUMOSAXReader.cpp
+++ b/src/utils/xml/SUMOSAXReader.cpp
@@ -24,6 +24,7 @@
 #include <config.h>
 
 #include <string>
+#include <memory>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


behrisch:
Thanks!

behrisch:
@jacobowitz  We want to add GPLv2+ as a secondary license to sumo and need your permission to also include this (and maybe code from other pull requests) in this license change. If you agree please add a +1 to #6506. This ticket also has more details and if you have questions you can ask them there. Thank you in advance!

PR #5934, opened 2019-08-14
Add optional parking entry/exit times set by vType parameter

theDivj:
Signed-off-by: theDiv <div@thoda.uk>

Code to optionally allow more realistic parking space entry/exit timings (as-is enters/leaves space in a single sim step) meaning lanes are blocked for longer than a single step for parking, obscuring parking related congestion.

Timings are set by a vType parameter of the form:
manoeuverAngleTimes="10 3.0 4.0,80 1.1 11.0,110 11.0 2.0,170 8.0 2.5,181 3.0 4.0"

Each triplet in the parameter consists of an angle value followed by an entry-time and exit-time. The angles delimit ranges corresponding to the angle between the space and the lane.

In the example above a space that is angled between 0 and 10 degrees from the lane will block the lane for 3 seconds before entering the parking space and for 4 seconds on leaving the space.

(2nd submission reflecting comments on pull #5894)

PR #5956, opened 2019-08-21
Add optional parking entry/exit times set by vType parameter

theDivj:

Code to optionally allow more realistic parking space entry/exit timings (as-is enters/leaves space in a single sim step) meaning lanes are blocked for longer than a single step for parking, obscuring parking related congestion.

Timings are set by a vType parameter of the form:
manoeuverAngleTimes="10 3.0 4.0,80 1.1 11.0,110 11.0 2.0,170 8.0 2.5,181 3.0 4.0"

Each triplet in the parameter consists of an angle value followed by an entry-time and exit-time. The angles delimit ranges corresponding to the angle between the space and the lane.

In the example above a space that is angled between 0 and 10 degrees from the lane will block the lane for 3 seconds before entering the parking space and for 4 seconds on leaving the space.

Code applies to <space> and roadside parking - with onRoad parking only the entry time value for zero degrees is applied because the vehicle is already in the lane.

(submission reflecting comments on pull #5894, #5934)


namdre:
Thanks again for the contribution!

behrisch:
@theDivj We want to add GPLv2+ as a secondary license to sumo and need your permission to also include this (and maybe code from other pull requests) in this license change. If you agree please add a +1 to #6506. This ticket also has more details and if you have questions you can ask them there. Thank you in advance!

PR #5979, opened 2019-08-22
Now it is possible to send parameters from LISA to SUMO

maxidigital:


behrisch:
Maxi is already working on a new version

PR #6009, opened 2019-08-28
SSM foe search and scan order

m-kro:
As mentioned in issue #5644 and #5527 , the current scan for foe candidates assigns the wrong conflict lane in some cases. This results in utterly underestimated TTC values. For more details look at the referenced issues. This pull request changes the search mechanism to remember visited lanes and look only at (possible) foe lanes in intersections.

Problems from the first pull request #5877 indicated by test cases have mostly been sorted out. It should be checked whether the failing tests currently have valid result files:

- decreased_extratime_following: in the result file ssm.sumo, extra time after conflict is longer than specified in the device
- foe_leader_deviating_routes: see above
- ticket5231: foe EW.6 should not be found using the specified range but has been in ssm.sumo

### DIFF
--- a/src/microsim/devices/MSDevice_SSM.cpp
+++ b/src/microsim/devices/MSDevice_SSM.cpp
@@ -177,27 +177,27 @@ std::ostream& operator<<(std::ostream& out, MSDevice_SSM::EncounterType type) {
 // static initialisation methods
 // ---------------------------------------------------------------------------
 
-std::set<MSDevice_SSM*>* MSDevice_SSM::instances = new std::set<MSDevice_SSM*>();
+std::set<MSDevice_SSM*, ComparatorNumericalIdLess>* MSDevice_SSM::myInstances = new std::set<MSDevice_SSM*, ComparatorNumericalIdLess>();
 
 std::set<std::string> MSDevice_SSM::createdOutputFiles;
 
 int MSDevice_SSM::issuedParameterWarnFlags = 0;
 
-const std::set<MSDevice_SSM*>&
+const std::set<MSDevice_SSM*, ComparatorNumericalIdLess>&
 MSDevice_SSM::getInstances() {
-    return *instances;
+    return *myInstances;
 }
 
 void
 MSDevice_SSM::cleanup() {
     // Close current encounters and flush conflicts to file for all existing devices
-    if (instances != nullptr) {
-        for (std::set<MSDevice_SSM*>::iterator ii = instances->begin(); ii != instances->end(); ++ii) {
-            (*ii)->resetEncounters();
-            (*ii)->flushConflicts(true);
-            (*ii)->flushGlobalMeasures();
+	if (myInstances != nullptr) {
+		for (MSDevice_SSM* device : *myInstances) {
+			device->resetEncounters();
+			device->flushConflicts(true);
+			device->flushGlobalMeasures();
         }
-        instances->clear();
+        myInstances->clear();
     }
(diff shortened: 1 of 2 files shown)


namdre:
* If you think the current state of this pull-request is "mostly" better than the current master, I'm inclined to merge the pull now despite some failing tests. 
* I'll take a look at ticket5231
* the current sumo c++ coding style uses no tabs. I can merge as it is but then i'll just modify most of your lines next time I save which somewhat clutters the authorship history.

m-kro:
Sorry for the tabs. Very bad many text editors hide them by default. I forgot to set my new installation of VS to replace them automatically. It has been corrected in the last commit.
For ticket5231, I forgot to mention the ego vehicle:

- at time 370, ego vehicle EW.16 is at pos 1762.57 of -center_0
- at the same time, foe vehicle EW.6 is at pos 1149.81 of -center_0 which is more than 200 away.

However then I still experience problems whith Texttest as my own test results also show a conflict with this vehicle pair but not in the single simulation run ?!

Yes, I would sustain to merge the request by now. And I might create a test with two intersections so that wrong conflict positions get recognized earlier in the development.

namdre:
For me it looks different at step 370:
(obtained on linux, fcd-output is the same regardless of running it from within texttest or manually).

```
<timestep time="370.00">
         ...
        <vehicle id="EW.16" x="560.75" y="11.61" angle="269.72" type="normal_car" speed="13.36" pos="2357.14" lane="-center_0" slope="0.00"/>
        <vehicle id="EW.6" x="357.52" y="10.63" angle="269.72" type="normal_car" speed="8.96" pos="2560.37" lane="-center_0" slope="0.00"/>
         ...
    </timestep>
```

schwamborn:
Tests decreased_extratime_following and foe_leader_deviating_routes are still failing:

```
TEST FAILED on ts-sim-build-ba : devices ssm config decreased_extratime_following

---------- Differences in ssm ----------
5c5
<     <conflict begin="1.00" end="10.60" ego="ego1" foe="foe1">
---
>     <conflict begin="1.00" end="10.00" ego="ego1" foe="foe1">

TEST FAILED on ts-sim-build-ba : devices ssm following foe_leader_deviating_routes

---------- Differences in ssm ----------
5c5
<     <conflict begin="1.00" end="14.60" ego="ego1" foe="foe1">
---
>     <conflict begin="1.00" end="14.00" ego="ego1" foe="foe1">
```

Are the new results as expected? If yes, I will approve them.

behrisch:
No, I think @m-kro is still working on this.

m-kro:
Well it has been some time I have been working on it. But the conflicts go into the end state at the same time, It's only that they are kept longer active afterwards in the old version which I could not track down yet. I'll try to understand the difference of the end time in the next days.

m-kro:
I cannot reproduce the texttest results with a single SUMO run at the moment, thus cannot debug it. Running SUMO directly gives me the same test results as those from the repo.

m-kro:
Sorry for the confusion about the SUMO versions. 
**I vote for approving the new results:** The difference in conflict durations is caused by the changed detection of foe vehicles inside junctions. Previously, all vehicles on a junction along the ego route where selected. Now, only those on the ego route or on internal edges marked as foes. Consequently, the conflict is marked as over when the foe vehicle enters the right turning edge while the ego vehicle continues straight. 
The geometry is somehow disregarded like with merging areas and should be fixed together in a separate topic.

PR #6046, opened 2019-09-07
Offset variable missing

rlloretb:
I had to add the variable offset in the function call. Else, all offsets were 0.


namdre:
Thank you!

PR #6116, opened 2019-09-25
Fixing mispell in GUI source code.

My5z0n:
Fixing doubled semicolon.
Signed-off-by: My5z0n <szmnnagel@gmail.com>


namdre:
Thanks. I'm surprised our regular astyle-runs did not catch this.

behrisch:
Just as a side note: astyle handles only braces and whitespace. @My5z0n Which tool did you use here or is this just a search and replace?

PR #6142, opened 2019-10-09
Fixes the traci variable value for crashed vehicles in the CC car-following model

thardes2:
The CC car following model for Veins-Plexe (Platooning) previously used the "getMyStops()" method to check if a vehicle cause a collision. 

This pull request changes the mechanism by considering the amount of collisions for a vehicle. If a vehicle has a `collsionCount>0` the according variable is set to true


namdre:
This commit changes semantics in an unreasonable way. The old version checks whether the vehicle under consideration is crashed at the moment. The new version checks whether any crashes have occured in the simulation at all (but those may have happened at some other time and to some other vehicle).

PR #6243, opened 2019-11-10
implement human readable time (using ISO 8601)

jetoff41:
for a better overview in the GPX output


behrisch:
Hi, I understand the intention and thanks for the PR, but this should probably rather be added to the output side (e.g. GPX) and not to the input, because this will now affect all outputs which have time. Furthermore you need to sign off your commit and agree to the Eclipse CDA if you did not do that yet, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work. I am closing this one since the changes need to be in a completely different file. Please open a new PR and/or come back for questions.

PR #6251, opened 2019-11-14
SSM device: check for invalid arrival times

m-kro:
Related to general topic #5527 .
For crossing conflicts, the TTC measure takes into account the estimated arrival and exit times of the vehicles in the conflict area. When a vehicle is at standstill, the function used for the time estimation returns INVALID_DOUBLE. Now this is checked before it leads to unreasonable TTC values.

### DIFF
--- a/src/microsim/devices/MSDevice_SSM.cpp
+++ b/src/microsim/devices/MSDevice_SSM.cpp
@@ -1288,7 +1288,7 @@ MSDevice_SSM::determineTTCandDRAC(EncounterApproachInfo& eInfo) const {
         if (myComputeDRAC) {
             drac = computeDRAC(eInfo);
         }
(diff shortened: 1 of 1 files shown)


namdre:
I'm reverting the patch because it breaks at lot of tests and to my understanding at least one of these is a regression.

Here is what I found:
- the INVALID value comes from estimateArrivalTime() which is used to compute the ConflictExitTime
- when a fast vehicle is braking, it may still cover enough distance to enter the conflict area but it will come to a standstill before leaving the conflict area (resulting in INVALID conflictExitTime)
- your additional check prevents computation of TTC in this case even though the vehicles could crash

Something else: when a vehicle is at standstill and part of a crossing conflict, computeTTC will be called with followerSpeed=0, leaderSpeed=0. This case appears to be handled and should not lead to unreasonable TTC values. Most likely I've missed somothing. Can you add a small example with unreasonable TTC output for our test suite?

PR #6260, opened 2019-11-15
avoid warning at compile time

andreastedile:
Otherwise, the following warning is thrown:
implicitly-declared ‘MSVehicle::Stop::Stop(const MSVehicle::Stop&)’ is deprecated [-Wdeprecated-copy]


behrisch:
Thanks for the PR! Since it is a single line only, I took the liberty of refactoring a little. We could not accept it as is because it is missing a signed off line, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work

PR #6316, opened 2019-11-25
Move Hello Sumo tutorial under Beginner Tutorials

PybloBenavides:
Hello Sumo should be under beginner instead of advanced tutorials I believe.
Thanks so much for the project!


namdre:
Thanks for the comment. After releasing the graphical network editor, the tutorial was moved to that section deliberately. (defining a network in XML is not a good first user experience). The intention was always to replace it with #6317

PybloBenavides:
Sounds good! Thanks a lot

Issue #6217, opened 2019-10-31
allow defining internal lane permissions


namdre:
setting internal lane permissions is possible via TraCI and respected by sumo since #6210

PR #6340, opened 2019-11-28
Define internal lane permissions

m-kro:
This pull request proposes a solution for #6217. Permissions are read/written for SUMO and plain XML formats. If the permissions have not been set explicitly in Netedit or specified in the plain XML or SUMO format, they should not be written to connection elements. The value SVC_UNSPECIFIED is used throughout the changes to reflect this situation and when building the network, the permissions of the successor lane are taken instead as it is now. 
The pull request completes the same netconvert tests successfully as the master branch currently does.

What the "PostProcessConnection" is used for? I have added the permissions there but am not sure they really are needed...


namdre:
Thanks!
Unfortunately this pull request has conflicts and isn't convered by the correct legal magic (the merge commit is to blame).
Can you refashion this PR into a fresh single commit with --signed-of?

PR #6354, opened 2019-12-02
Accept generic parameter `lcOvertakeRight` in LC2013

kschrab:
The lane change parameter `lcOvertakeRight` could not be changed via TraCI due to a missing case in the methods `getParameter` and `setParameter` in MSLCM_LC2013.


namdre:
Thanks for the pull request. The header also had to be changed or it would not compile due to const-ness. (fixed via 2ff295360db8a78082ff1b2b1c62e297739bfd09)

PR #6401, opened 2019-12-11
Squashed commit of ElectricHybrid

jprk:
@namdre I hope I did not break anything in you latest commits as one compilation check before the commit into a pull request branch takes ages for me and of course you managed to commit some changes into master inbetween ;-).

The PR contains the code for:
* elecHybrid device,
* traction substation,
* overhead wires including GUI
* overhead wire circuit representation,
* optional circuit solver.

It further contains:
* initial documentation,
* two very basic tests,
* XSD updates for new tags that we added.

Requires [Eigen](https://eigen.tuxfamily.org/) >= 3.2 for the circuit solver to work.  For Windows, Eigen 3.3.7 is part of SUMOLibraries, on Linux systems it should be available as a package (at least on Ubuntu it seems to be so).

Currently implement only a very limited mesoscopic functionality, we would like to fist stabilise this and then apply the mesoscopic patch over the stable microscopic code.

Jan & Jakub

Signed-off-by: Jan Přikryl <prikryl@rice.zcu.cz>


namdre:
After replacing isnan (`error: call of overloaded ‘isnan(double&)’ is ambiguous`) with our own ISNAN macro, I could successfully compile the patch. yay!

jprk:
Yessss! Thanks a lot.

namdre:
@jprk I hope the changed option names don't cause too much inconvenience.

behrisch:
@jprk I just noticed that thew tests sometimes cause NAN output, see https://sumo.dlr.de/daily/msvc12extrax64report/sumo/test_default.html#extra Do you have plans to change that? (We would prefer an explicit "NA" as with the SSM device output.)

jprk:
I will look into that. I suppose I used the standard double to string
conversion routines, I will check the SSM code to see how the NA output is
being generated.

Jan

Dne po 20. 1. 2020 8:58 uživatel Michael Behrisch <notifications@github.com>
napsal:

PR #6425, opened 2019-12-17
Fix version.h problems for Linux

atellyro:
Signed-off-by: Harald <fechsaer@gmail.com>

### DIFF
--- a/tools/sumolib/version.py
+++ b/tools/sumolib/version.py
@@ -25,15 +25,22 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 import subprocess
+import os
 from os.path import dirname, exists, join
 
 UNKNOWN_REVISION = "UNKNOWN"
 GITDIR = join(dirname(__file__), '..', '..', '.git')
+verbose = False  # set to True if you want some more debug statements
 
 
 def _findVersion():
+    if verbose:
+      print('gitdir1 = %s' % GITDIR)
+      print('pwd1 = %s'    % os.getcwd())
     # try to find the version in the config.h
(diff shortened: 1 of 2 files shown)


behrisch:
Thank you for the PR! Is this only to fix things when you have an installed and a developer version?

tools/sumolib/version.py
```diff
@@ -25,15 +25,22 @@
...
 
 
 def _findVersion():
+    if verbose:
+      print('gitdir1 = %s' % GITDIR)
+      print('pwd1 = %s'    % os.getcwd())
     # try to find the version in the config.h
     versionFile = join(dirname(__file__), '..', '..', 'include', 'version.h')
+    if not exists(versionFile):
+      versionFile = join('src', 'version.h')
```
behrisch: There should be no version.h in src anymore. This is an artifact created only by "in tree" builds which are either using autotools or a discouraged cmake configuration.

atellyro:
This script runs with pwd = .../build/cmake-build
So it picks up the version file generated in the build dir

Am 18.12.19 um 09:26 schrieb Michael Behrisch:

atellyro:
Am 18.12.19 um 09:23 schrieb Michael Behrisch:
>
> Thank you for the PR! Is this only to fix things when you have an 
> installed and a developer version?
>
Yes,  except:

   you do some structural changes to sumolib.version.  In this case the 
eggs are called first, independent from PYTHONPATH etc.

   in this case you have to remove the installed sumolib

behrisch:
thanks for the PR. I tried to solve the problems without adding too mucg debugging code. Please review whether your expectations are met.

PR #6431, opened 2019-12-19
SAGA added to the Contributed docs. 

lcodeca:


lcodeca:
I cannot find a way to sign my commits from the GitHub interface.
I'm going to try again another time.

behrisch:
It works for me, if I manually add an empty line and the signed-off footer in the commit description field.  But since you have commit rights anyway it is not that important.

lcodeca:
I don't know why I cannot. You can see it happening in https://github.com/eclipse/sumo/pull/6435 too.
Apologies for this mess.

behrisch:
My GitHub gives me "which email address associate with the commit" dialog field where you probably need to select a mail address fitting the signed off?

Issue #6147, opened 2019-10-11
add new traci domains


namdre:
Currently we use 15 domains - last 4 bits of the command (with one left for expansion)
The first 4 bits are used to distinguish 
- CMD_SUBSCRIBE_XXX_CONTEXT
- RESPONSE_SUBSCRIBE_XXX_CONTEXT
- CMD_GET_XXX_VARIABLE
- RESPONSE_GET_XXX_VARIABLE
- CMD_SET_XXX_VARIABLE
- CMD_SUBSCRIBE_XXX_VARIABLE
- RESPONSE_SUBSCRIBE_XXX_VARIABLE

This leave one unused bit which we can use to double the available domains.
We could add:

- [x] busstop   
- [x] parkingarea
- [x] chargingstation
- [x] routeprobe
- [x] calibrator
- [x] rerouter
- [x] variablespeedsign
- [x] meandata
- [x] overheadwire

Still missing:
- [ ] subscriptions (context and otherwise)

namdre:
API questions
- detectors (routeProbe, meanData) aggregated data and could either retrieve the (complete) last data interval or the (incomplete) current interval. Either
  - add two methods getLastXXX, getCurrentXXX  (pro: better for subscriptions, con: more methods)
   - add one method with extra parameter getXXX(..., bool current)

- meandata return lots of things 
  - add new method per returned property
  - add one method that returns a struct
  - retrieve data for individual edges
  - retrieve list data for all edges

- routeProbe and vss have paired data (route, count), (time, speed)
   - add two methods (getRouteIDs, getRouteCounts)
   - add one method two return list of structs for encodng the pair

behrisch:
subscriptions still open, separate tickets for the remaining parts https://github.com/eclipse/sumo/issues/6147#issuecomment-675320599

behrisch:
Subscriptions have more or less been solved in 0b5399a4b5de15122746eca184ac7a2341b61bbc

PR #6434, opened 2019-12-19
Added retrieval of a bus stop's position in its lane

tkchouaki:
Hi,

I added the ability to retrieve the bus stop's begin position & and position on its lane via libsumo (and TraCI).

Regards.


namdre:
Thanks for the pull request (and sorry for the late reply). I'm reluctant to merge it because our goal is to add access to all the stopping place functionality via #6147. Any more calls that we add to traci.simulation in the meantime would then duplicate functionality or break compatibility if we do not keep them.

PR #6508, opened 2020-01-19
Cosmetic changes to Readme.md

claell:
I made some cosmetic changes to the Readme, including fix of a few typos and using code formatting for options. Let me know whether those changes do not fit your formatting style.

### DIFF
--- a/docs/web/docs/FAQ.md
+++ b/docs/web/docs/FAQ.md
@@ -96,7 +96,7 @@ git commit -m "your change message" -s
 ```
 
 !!! caution
-    the option -s must be used so the contribution can be accepted in accordance with Eclipse rules
+    the option `-s` must be used so the contribution can be accepted in accordance with Eclipse rules
 
 - Push your changes to your fork on GitHub with
 
@@ -186,7 +186,7 @@ always use the latest version of SUMO.
 ### What should I do to get helpful answers on the mailing list?
 
 - Make your question specific
-  - avoid vague terms.
+  - Avoid vague terms.
   - Always name the SUMO version to which your question applies
   - Include relevant warnings/errors/stack-traces in your question
     (please copy the text and avoid screenshots to reproduce text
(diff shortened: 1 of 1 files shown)


behrisch:
Thanks!

namdre:
actually, I prefer to distinguish `xml input stuff` from **--option-stuff**. The whole changelog is written in this style. Please don't do more pull requests for this particular cosmetic change before we've discussed it internally.

claell:
@namdre I noticed that you mostly use the bold formatting for command line options. Thus I wasn't sure, whether the style change is ok. Apparently, at least in the changelog, you use the code formatting mostly, but not exclusively for XML input (I noticed some instances where it was used for parameters for command line options, I guess). However, the usage differs in other places of the docs, while at most times options seem to be formatted in bold.

To me it seemed nicer to have such options also formatted as code (at least that is how many other projects tend to do it).

behrisch:
We discussed this and decided to stay with the old style for now. Please feel free to do another pull request applying this style consistently. Furthermore having another look at the changes I am not sure what the correct English way of dealing with bullet point lists is. In German you only start with a capital letter when it is a complete sentence.

claell:
Alright.

> applying this style consistently

So you would approve a change of the style if it is overall consistent? I might have a look again on how other large projects do their formatting and then change it altogether if I find the time.

> the correct English way of dealing with bullet point lists is

According to what I have found when I looked that up there is not just one correct way, but can also depend on the style guideline in question. However, most seem to have in common that they capitalize whole sentences which then end with a dot. When the bullet point list is rather a enumeration that could also be a whole sentence altogether, the points are seperated by comma and a colon at the end of the last point. The points are not capitalized in that case, if not needed.

For this PR, I just adjusted the points that did not match the overall formatting, to establish consistency.

> In German you only start with a capital letter when it is a complete sentence.

As a rule of thumb that seems to be correct, yes: https://federwerk.de/aufzaehlungen-gross-und-kleinschreibung-und-interpunktion/

namdre:
> Alright.
> 
> > applying this style consistently
> 
To clarify, we decided on the style 
- options in bold (**--some-option**)
- xml snippets as code (use atttribute `foo` to configure feature XY)

As this style is not yet used everywhere in our documentation, you are very welcome to help by changing options to bold and xml snippets to code style.

claell:
Thanks for clarifying. Since I don't really agree to this formatting without further explanation, I probably won't change it in missing places. Just an example how others use code formatting (Python doc): https://docs.python.org/3/using/cmdline.html#command-line, https://docs.python.org/3/installing/index.html

PR #6509, opened 2020-01-19
Add some more information to OSMWebWizard.md

claell:
When going through the tutorial I cam across two more complicated points that were not mentioned, so I added them.


behrisch:
SUMO_HOME should be set automatically. Which ubuntu package did you use?

claell:
I used the one from the PPA as was found in the docs.

claell:
Problem is maybe that I did not log out before invoking the script.

behrisch:
So additions seem to be unnecessary, if not please reopen.

Issue #6506, opened 2020-01-17
add GPL v2.0+ as secondary license


behrisch:
This ticket tracks the progress in adapting SUMO's license from a pure Eclipse Public License v2 to one which uses the Exhibit A to allow people to license SUMO under GPL v2 or later as well. For details see https://www.eclipse.org/legal/epl-2.0/faq.php#epl-secondary. We need every contributor and committer to add a +1 to this ticket. Please do not use the GitHub "reactions" to send a  simple "thumbsup" but post a comment to this ticket with at least a +1 in it (for an example see below). You can also use the comments to ask for clarifications. Thanks for your support!

behrisch:
+1

angelobanse:
+1

RobertHilbrich:
+1

lbieker:
+1

ArmelliniMG:
+1

schwamborn:
+1

namdre:
+1

palvarezlopez:
+1

m-kro:
+1

michele-segata:
+1

lcodeca:
+1

jprk:
+1

rummel123:
+1

yunpangfloetteroed:
+1

RNippold:
+1

mightyCelu:
+1

theDivj:
+1

seb-ehm:
+1

bstriner:
+1

srishtidh:
+1

tkchouaki:
+1

mkrumnow:
+1

sommer:
+1

gprichter:
+1

dbuse:
+1

Anisotrop:
+1

jacobowitz:
+1

behrisch:
Thank you all! The license change is now complete. The official announcement will be made once the Eclipse process of certifying the 1.5.0 release is through.

PR #6580, opened 2020-01-30
Patches osmGet.py to use proxy authentication.

Anisotrop:
Behind a corporate proxy with authentication the script could not create the tunnel.


behrisch:
Thanks for the PR! Could you please use the standard indentation style of 4 spaces.

namdre:
also, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work
and #6506

behrisch:
@Anisotrop Thanks again. Please add a +1 here #6506, so that we can use your code after our license adaptation.

Issue #6642, opened 2020-02-12
Lanearea detector: jam length exceeds detector length


m-kro:
Currently, the detector can output values exceeding its own length due to how the first vehicle is treated:
https://github.com/eclipse/sumo/blob/2ac57f0e4cbafd56ad8fbd4df02c60d6e962174d/src/microsim/output/MSE2Collector.cpp#L1199-L1201
"distToDetectorEnd" gets negative when the vehicle front has already left the detector but still contributes to the result. **What is the intended semantics of the lanearea jam length?** I suggest 
`
const double jamLengthInMeters = lastVeh->distToDetectorEnd - MAX2(0.,firstVeh->distToDetectorEnd) + lastVeh->lengthOnDetector;
`
to limit the jam length to the detector length. If you want I can prepare a pull request for this.

namdre:
I also think that the jam length should be limited to the detector length. Your opinion @behrisch?

behrisch:
It is definitely wrong the way it is but I am unsure whether one would expect here the sum of the length of all vehicles plus gaps. m-kro's  solution has the side effect that for a long vehicle and a short detector the jam length could be less than the vehicle length. But still better than before ;-), so yes please do a pull request unless you find my remark so convincing that you want to introduce a configurable version of that :-)

PR #6661, opened 2020-02-16
limit lanearea jam to detector length

m-kro:
As explained in #6642 , lanearea detectors should only return jam lengths which do not exceed the detected area. Before this PR, this has been respected only for the last but not for the first vehicle.


namdre:
this PR was causing tests/sumo/output/e2multilane/teleportOutOfDetector to return negative jamLength. (There is a vehicle that has occupies the detector with is back part and the jam length correctly registered only the overlap length befre). The alterantive solution is more explicit.

Issue #3552, opened 2017-11-01
TraCI induction loop considers vehicles of the current time step


behrisch:
it should only consider values of the last time step as stated in documentation and function names

aminissn:
Hi Michael,
do you have any idea when this issue will be solved? I guess it is still unsolved for version 32!

Thanks!
Sasan

behrisch:
No sorry, but I put it on my list.

aminissn:
Is there a workaround for this bug? One other thing that can be nice to have is to get occupancy=100% when a non-moving vehicle stands on the detector. Currently, I get -1. I guess it gives values when getTimeSinceDetection is 0, which is not the case when there is congestion on the detector.

behrisch:
@aminissn Could you test this or even better send a small test scenario?

aminissn:
Here is the example I was testing with. I am basically comparing the FD from TraCI and loop detector output file. 
[01_onelane.zip](https://github.com/eclipse/sumo/files/3845413/01_onelane.zip)

m-kro:
@behrisch: Due to your changes at [https://github.com/eclipse/sumo/blob/c31e7cbd20acffe7a4b9a0e5365190993351b2b3/src/microsim/output/MSInductLoop.cpp#L295](url), standing vehicles are only considered in one time step for e.g. TraCI inductionloop.getLastStepVehicleNumber. Only when it just has entered the detector, its enter time fulfills the condition. Is that really the desired behaviour?

behrisch:
@m-kro No, it is not. It is very high on my ToDo-List for the new year to solve this ticket. My general approach would be to let the behavior mimic a real loop as close as possible, meaning all values which require that the vehicle has already left the loop (speed, length, ...) are only available when the vehicle has left. Everything else should move to the instant loop. @aminissn Do you agree here?

aminissn:
@behrisch I agree on making the values available after the vehicle has left the detector.

aminissn:
I totally forgot about this matter and was trying to implement an ALINEA controller and noticed the difference between the Fundamental Diagram obtained from TraCI and the generated E1 detector output. Just curios about the best way to get the "right" occupancy and flow measurements using TraCI for actuated control applications?
![image](https://user-images.githubusercontent.com/30176885/153222273-49bfd236-d0b1-42c4-a144-7508db9687e4.png)

Issue #3352, opened 2017-08-15
allow rerouting of pedestrians (trac #3352)


behrisch:
Migrated from http://sumo.dlr.de/ticket/3352
```json
{
    "status": "closed", 
    "changetime": "2017-08-15T21:02:17Z", 
    "description": "", 
    "reporter": "namdre", 
    "cc": "", 
    "resolution": "fixed", 
    "_ts": "1502830937532231", 
    "component": "TraCI", 
    "summary": "allow rerouting of pedestrians", 
    "priority": "major", 
    "keywords": "", 
    "time": "2017-08-15T10:50:53Z", 
    "milestone": "1.0.0", 
    "owner": "namdre", 
    "type": "enhancement"
}
```

PR #6682, opened 2020-02-20
Consider arrived vehicles for induction loop occupancy

m-kro:
Due to ongoing refactoring and semantic changes/corrections of the induction loop (see #3552), the detector occupancy does not consider vehicles which have already arrived on the detector more than one time step ago. This has a severe impact on any application relying on the occupancy data, e.g. external signal control, as at most times 0 will be returned.

@behrisch I understand that you are working on it. Nonetheless, occupancy output is a very basic and rather "old" feature and probably well used. Please think of accepting the pull request as long as there is no long-term solution.

### DIFF
--- a/src/microsim/output/MSInductLoop.cpp
+++ b/src/microsim/output/MSInductLoop.cpp
@@ -169,7 +169,7 @@ MSInductLoop::getOccupancy() const {
     const SUMOTime tbeg = SIMSTEP - DELTA_T;
     double occupancy = 0;
     const double csecond = SIMTIME;
-    for (const VehicleData& i : collectVehiclesOnDet(tbeg)) {
+    for (const VehicleData& i : collectVehiclesOnDet(tbeg, false, true)) {
         const double leaveTime = i.leaveTimeM == HAS_NOT_LEFT_DETECTOR ? csecond : MIN2(i.leaveTimeM, csecond);
         const double entryTime = MAX2(i.entryTimeM, STEPS2TIME(tbeg));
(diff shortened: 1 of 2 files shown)


behrisch:
@m-kro Could you please recheck whether the recent changes made this one obsolete?

m-kro:
You changed other things but left the one important line untouched:
https://github.com/eclipse/sumo/blob/4998a1791f235fccebd5b3a7a294502c8a1481ed/src/microsim/output/MSInductLoop.cpp#L291 
This means vehicles standing on the detector for more than one time step have no impact on detector occupancy because they arrived earlier than "t". This way, only rising edges ("steigende Flanke") can be observed at the right time. I'm unsure whether this is intended when you write in #3352  

> meaning all values which require that the vehicle has already left the loop (speed, length, ...) are only available when the vehicle has left. **Everything else should move to the instant loop**

For the moment I have replaced my TraCI occupancy calls with traci.inductionloop.getTimeSinceDetection() as this still considers vehicles on the detector without extra conditions. Could you clarify what you want detector occupancy to be...?

behrisch:
I implemented this now but it needs more tests. @m-kro We also discussed the topic internally whether a vehicle which changes the lane (or arrives / teleports) while being on the loop should be counted. The intermediate result was, that it is practically impossible to leave the lane "sideways" so this is just an artifact of sumo's instant lane changing and the vehicle should still count. Do you agree here?

m-kro:
@behrisch Yes I agree.

PR #6709, opened 2020-02-28

theDivj:

Introduces an 'idling' state to allow capture of emissions from vehicles waiting to enter the net and add this blocked time as waiting time.  (Issue 2233)

Changes to support coloring by emissions of parked and idling vehicles to be added

As-is emission models have these reconciliation discrepancies which have not been addressed in these changes:
1. emission output does not reconcile with tripinfo because tripinfo 'loses' the first emission value for each vehicle and the emissions output 'loses' the last value for each vehicle.
2. tripinfo and edge/lane emissions sometimes fail to reconcile because edge/lane emissions apply a scaling value when the last timestep is shorter than TS.


theDivj:
silly error - might have expected visual studio to have caught it though :(

PR #6710, opened 2020-02-29

theDivj:

Introduces an 'idling' state to allow capture of emissions from vehicles waiting to enter the net and add this blocked time as waiting time. (Issue 2233)

As-is emission models have these reconciliation discrepancies which have not been addressed in these changes:
a) Emission output does not reconcile with tripinfo because tripinfo 'loses' the first emission value for each vehicle and the emissions output 'loses' the last value for each vehicle.

b) tripinfo and edge/lane emissions sometimes fail to reconcile because edge/lane emissions apply a scaling value when the last timestep is shorter than TS.

Changes to GUIVehicle.cpp to follow - fixing coloring by emission of parking/idling vehicles.

### DIFF
--- a/src/mesosim/MEVehicle.cpp
+++ b/src/mesosim/MEVehicle.cpp
@@ -166,6 +166,10 @@ MEVehicle::isOnRoad() const {
     return getSegment() != nullptr;
 }
 
+bool
+MEVehicle::isIdling() const {
+    return false;
+}
 
 bool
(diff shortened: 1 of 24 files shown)


theDivj:
This pull confirming all builds - some additional updates to unrelated test outputs are needed (because of tripinfo waiting time increases) so I'll close this pull if it passes all tests then submit the complete set

theDivj:
done

PR #6711, opened 2020-02-29

theDivj:

Introduces an 'idling' state to allow capture of emissions from vehicles waiting to enter the net and add this blocked time as waiting time. (Issue 2233)

As-is emission models have the following reconciliation discrepancies which have not been addressed in these changes:
a) Emission output does not reconcile with tripinfo because tripinfo 'loses' the first emission value for each vehicle and the emissions output 'loses' the last value for each vehicle.

b) tripinfo and edge/lane emissions sometimes fail to reconcile because edge/lane emissions apply a scaling value when the last timestep is shorter than TS.

Changes to GUIVehicle.cpp to follow - fixing coloring by emission of parking/idling vehicles.


theDivj:
Have just realised that changes to GUIVehicle are not required - the changes to MSBaseVehicle achieve the same effect. (Correcting emission coloring of parked/idling vehicles)

PR #6768, opened 2020-03-09
Fix a bug causing vehicles to not be untracked in GUI

NyghWolf:
Signed-off-by: Farkas Tamás Gábor <farkastamas1095@gmail.com>


namdre:
Thanks!

PR #6822, opened 2020-03-18
Do not return the leader if it is further than the required distance

andreastedile:
When the distance from the ego vehicle to its leader is greater than the desired distance, the pair (0, -1) should be returned. Currently, the leader is returned regardless of its distance.

Signed-off-by: andreastedile <andrea.stedile@studenti.unitn.it>

### DIFF
--- a/src/microsim/MSVehicle.cpp
+++ b/src/microsim/MSVehicle.cpp
@@ -5435,7 +5435,6 @@ MSVehicle::getDistanceToPosition(double destPos, const MSEdge* destEdge) const {
     return distance;
 }
 
-
 std::pair<const MSVehicle* const, double>
 MSVehicle::getLeader(double dist) const {
     if (myLane == nullptr) {
@@ -5453,8 +5452,12 @@ MSVehicle::getLeader(double dist) const {
         lead = *(it + 1);
     }
     if (lead != nullptr) {
-        std::pair<const MSVehicle* const, double> result(
-            lead, lead->getBackPositionOnLane(myLane) - getPositionOnLane() - getVehicleType().getMinGap());
+        double gap = lead->getBackPositionOnLane(myLane) - getPositionOnLane() - getVehicleType().getMinGap();
+        if (gap > dist) {
+            lead = nullptr;
+            gap = -1;
+        }
+        std::pair<const MSVehicle* const, double> result(lead, gap);
         lane->releaseVehicles();
         return result;
     }


namdre:
In my understanding, the current behavior is intended. The distance feature is mostly meant as a performance saver (to avoid searching the whole road network when there are only a few vehicles in it). Since the distance is returned as part of the result, the end user can still filter on that value if a hard cut-off is desired.

@behrisch what is your opinion on this?

andreastedile:
Let me reiterate that the function specification is "Returns the leader of the vehicle looking for a fixed distance". Right now, if the leader is 10 meters further and I pass a distance of 5 meters, it still gets returned. In my opinion, this is a bit counter-intuitive.

behrisch:
I agree to @namdre . I will fix the documentation.

andreastedile:
I have tested the function again. If I spawn two vehicles with an inter-vehicle distance of 10 meters and let the second one request its leader with a minimum look ahead distance greater than 10 meters, the first vehicle gets returned.
Is this how it is supposed to work? Shouldn't such look ahead distance make the function skip the first vehicle entirely?

behrisch:
I am not sure, if I understand your scenario correctly. When asked about vehicle v2, the function should return the first vehicle downstream v2's current route (more precisely the list of its current best lanes) and abort the search not earlier than the given distance. If v1 is downstream the route of v2 it is correct. As namdre mentioned the distance is merely an optimization thing to avoid large lookaheads.

andreastedile:
Thanks for your clarification, I think that now I get what you mean. I wrongly thought that the value in question was the minimum distance from which the research should _start_; instead, it is the distance after which the research can be aborted.
Now, suppose a scenario in which two vehicles proceed on the same lane in a stretch of freeway, and are 150 meters apart. What value of look ahead ensures that, when the second vehicle (the one behind) asks for its leader, will _effectively_ get the leader? Clearly, 160 meters should be enough, because the research won't stop before this distance, and the preceding vehicle falls before this range.
Is there a thumb rule for this, that is also valid for other scenarios?
Let me know if my question wasn't clear enough.

namdre:
You should set the distance value to the maximum range at which the leader vehicle is still relevant to you. (i.e. you might find a leader at 5000m distance but that is not interesting to you). 
One way to look at it is in terms of braking distance (if the leader was a stationary obstacle, it would obviously be relevant when within the braking distance of the ego vehicle).  This means you could take the braking distance at the maximum speed of the ego vehicle as a boundary for the leader search.

andreastedile:
Perfect answer: thank you.

PR #6841, opened 2020-03-28
improve doxygen output and explain it

atellyro:
The sumo.doxyconf is changed to:
- do not use UML LOOK (makes diagrams easier to read)
- use only 3 columns in data structure index (all columns fit on one page)
- create a caller graph
- all graphs shows only the next 2 neighbors (much easier to read)
Then there is a new doc page to explain the working with doxygen output.
Many people wants to explore e.g. CF-Models
Feel free to adapt the doc page!


atellyro:
I have merged the old changes to the current main branch

behrisch:
Thank You! I merged just 6259018867b9637870dff91c2aea91c91930b912 right now.

PR #6950, opened 2020-04-28
Fixed typos

repa1030:


namdre:
Thanks for the correction. (I could not apply this directly due to ECA).

PR #7080, opened 2020-05-27
add how to add multiple vTypes by running randomTrips.py multiple times

DayuanTan:
Signed-off-by: BigCircle <zhuceafter2015@163.com>

I added a subsection about how to include multiple vTypes in a simulation by running randomTrips.py to generate moultiple route files and feed them into sumo at the same time.

I think it will be helpful for people who want to implement the vType distribution by hand.

### DIFF
--- a/docs/web/docs/Definition_of_Vehicles,_Vehicle_Types,_and_Routes.md
+++ b/docs/web/docs/Definition_of_Vehicles,_Vehicle_Types,_and_Routes.md
@@ -889,6 +889,29 @@ following:
 </routes>
 ```
 
+### Adding multiple vTypes by hand
+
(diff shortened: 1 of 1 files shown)


namdre:
Thank you for the pull request. I do not wish to add it in the suggested section because that page is for the raw specification rather than tool-specific advice.
I think it would be worthwhile to add this either
- in a subsection of the randomTrips documentation
- in a separate tutorial page ("Building a randomTrips scenario with different vehicle types"). As a tutorial page it would also be fitting to add both variants we discussed (multiple calls vs. vTypeDistribution).

I'd be happy to accept another pull request for either of these variants.

PR #7086, opened 2020-05-27
createVehTypeDistribution to allow emissionsClass with numeric characters 

mschrader15:
When listing an emissions class such as 'HBEFA3/PC_G_EU4' in the parameter.txt file for createVehTypeDistribution.py, the numeric characters caused the argument to be marked as numeric.

Caused an error at Line 55: https://github.com/mschrader15/sumo/blob/b9b859b59956d4a408eac5ecb51144e371effe66/tools/createVehTypeDistribution.py#L55

Signed-off-by: Maxwell Schrader <mcschrader@crimson.ua.edu>

### DIFF
--- a/tools/createVehTypeDistribution.py
+++ b/tools/createVehTypeDistribution.py
@@ -228,7 +228,10 @@ def readConfigFile(options):
                             break
 
                     if not distFound:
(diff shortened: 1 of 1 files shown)


behrisch:
Thanks!

PR #7102, opened 2020-06-02
Better float check for createVehDistribution.py

m-kro:
Currently, other values starting with numbers would be qualified as numeric, even if they aren't. As spaces are stripped before the type check, we can use the regex start/end notation to avoid misclassified data.


behrisch:
Wouldn't it be easier to try converting to float and catch the exception (and arguably more pythonic too)? This would avoid the regex and support scientific notation.

PR #7290, opened 2020-07-07
PyPML, SAGA and RLLIB+SUMO Utils projects added to the list.

lcodeca:
Signed-off-by: Lara CODECA <lara.codeca@gmail.com>


behrisch:
more projects more fun, thanks Lara!

PR #7292, opened 2020-07-07
Fix "static variable should be qualified by type name" javac warnings

mernst:
When compiling traas, javac issues 4 errors of the form

```
[static] static variable should be qualified by type name, Config, instead of by an expression
```

This pull request corrects them.

### DIFF
--- a/tools/contributed/traas/src/main/java/de/tudresden/sumo/util/Sumo.java
+++ b/tools/contributed/traas/src/main/java/de/tudresden/sumo/util/Sumo.java
@@ -131,7 +131,7 @@ public class Sumo {
     public void start_ws() {
 
         //start SUMO
-        conn = new SumoTraciConnection(conf.sumo_bin, conf.config_file);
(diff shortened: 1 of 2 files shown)


behrisch:
Thanks for the PR! Unfortunately you have not filled in the ECA or the checker did not recognize it. Please see https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement for details. Could you still do this?

mernst:
@behrisch Thanks for checking in.

I followed the link and filled out the form. and [the commit](https://github.com/eclipse/sumo/pull/7292/commits/99d4947c576e7b73024067a51fc4e434b42171bb) has "Signed-off-by:" text.
I did follow the link after I created the pull request (that's when the eclipsefdn/eca check failed and I got the link).

Is it possible that the check is out of date since it was done at the moment of the pull requset?
Or, can you tell me what else I need to do?
Thanks.

behrisch:
The problem was exactly as you mentioned. Now everything is fine. Thanks again!

Issue #7190, opened 2020-06-18
Adding a TraCI-command


Domsall:
To create a submicroscopic simulation, I have combined a virtual vehicle in Simulink with SUMO via the C++ TraCI-API (or using TraCI4Matlab).
Many different TraCI-commands already exist for achieving such tasks: e.g. with getLeader, getNextTLS, getBestLanes, and so on, one can retrieve the variables needed for a CF/LC-model working outside of SUMO.
Another way of communicating with SUMO is to work with the CF-models already implemented in SUMO via commands like getfollowSpeed and getStopSpeed.
In my opinion, it is a hard task to create own driver models for the submicroscopic simulation with SUMO, which should work similar to the included CF/LC/junction-models.

Back in my SUMO v1.0.1 fork, I added a TraCI-command which lets the Simulink model benefit from the extensive model library of SUMO.
Now I would like to know if it makes sence to push this command back to the newest SUMO version in the main branch (or if such tasks "SUMO<->virtual vehicle" are normally handled differently).
The command is called get/setFutureSpeed and works similar to get/setSpeed. When using the command setFutureSpeed, the SUMO-vehicle is put into a remote listening mode.
In every time step and for every vehicle, SUMO calculates vNext. Instead of now updating the ego-vehicle with vNext, the intended acceleration is sent to the Simulink-vehicle, which tries to reach vNext.
As the Simulink-vehicle will never perfectly reach this velocity, the actual achieved speed and position is sent back to SUMO (moveTo, setFutureSpeed). SUMO then calculates the next vNext with the new position and speed.
The advantage of this operation mode is that the user can use the different models of SUMO (e. g. ACC, CACC, IDM) and benefit from further development without changing the control concept of the Simulink vehicle.
Overriding vNext of SUMO is still possible (e. g. when transitioning between automated and conventional driving or to force a lane change).

namdre:
If I understand you correctly, you need an API to modify the speed computed by the SUMO models in accordance with an external model of how a physical vehicle reacts (engine response etc..).
To do so you need "intendedAcceleration" and this should affect the observed vehicle acceleration, speed and position.

The existing commands are not well suited for this:
- moveToXY requires x,y positions rather than accelerations/speed (it does however updated speed along with the position)
- moveTo only corrects the lane-position but does not correct speed/accelration and also requires knowledge of the road topology when adapting speeds at edge boundaries
- setSpeed sets a fixed speed for future time steps rather than doing a one-time correction

[...]

PR #7294, opened 2020-07-08
Adding a TraCI-command setPreviousSpeed, see issue #7190

Domsall:
I also added the TraCI-command setMinGap for the Vehicle-Class.
I used 0x3c, as I used to have 0x38 for this command, which was taken by now.
This command isn't a fancy implementation, as it only changes mySpeed, but as a few tests showed, it worked good.

### DIFF
--- a/src/libsumo/TraCIConstants.h
+++ b/src/libsumo/TraCIConstants.h
@@ -832,6 +832,9 @@ TRACI_CONST int VAR_SLOPE = 0x36;
 // speed (get: vehicle)
 TRACI_CONST int VAR_SPEED = 0x40;
 
+// adapt previous speed (set: vehicle)
+TRACI_CONST int VAR_PREV_SPEED = 0x3c;
+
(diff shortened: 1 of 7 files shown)


behrisch:
Can you please give a short description of your use case? I am not sure where this is useful.

Domsall:
As pointed out in issue #7190 , I run SUMO with Simulink and therefor use a command to modify the speed of an ego-vehicle in SUMO before the next time step. The intended speed is calculated with the CF/LC-model and the modified speed only holds for one update/time step. The method works like this:

- Simstep
- getSpeed/Acceleration from SUMO
- try to reach that Speed/Acceleration in Simulink vehicle model
- can't perfectly reach it, so update SUMO with reached Speed (setPreviousSpeed) and updated position (moveTo/moveToXY)
- Simstep

behrisch:
Thanks a lot!

PR #7309, opened 2020-07-13
fix(matsim_importPlans.py): skip persons with zero plans

asiripanich:
Feel free to merge if you think this is useful. I have some people with zero plans in my MATSim plan file which raises `IndexError: list index out of range` error.


behrisch:
Thanks for the PR. I applied it in a slightly different way (could not accept it because the paperwork seems to be incomplete, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work). Thanks anyway!

PR #7314, opened 2020-07-14
Change some SumoCommand formal parameters from Object to int

mernst:
`new SumoCommand` is always passed int values for certain of its parameters.  The int values get boxed up into an Integer, upcasted to Object, casted back down to Integer, then unboxed to int.  This pull request changes the formal parameter type to `int`, which improves run-time efficiency.  More importantly, it makes the code clearer and enables mistakes to be caught at compile time rather than at run time.

### DIFF
--- a/tools/contributed/traas/src/main/java/de/tudresden/sumo/util/SumoCommand.java
+++ b/tools/contributed/traas/src/main/java/de/tudresden/sumo/util/SumoCommand.java
@@ -54,18 +54,18 @@ public class SumoCommand {
     int output_type;
 
     //Get Statements
-    public SumoCommand(Object input1, Object input2, Object input3, Object response, Object output_type) {
+    public SumoCommand(int input1, int input2, Object input3, int response, int output_type) {
 
-        this.cmd = new Command((Integer) input1);
-        cmd.content().writeUnsignedByte((Integer) input2);
+        this.cmd = new Command(input1);
+        cmd.content().writeUnsignedByte(input2);
         cmd.content().writeStringASCII(String.valueOf(input3));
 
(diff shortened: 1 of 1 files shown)


mernst:
Could someone take a look at this pull request?

tools/contributed/traas/src/main/java/de/tudresden/sumo/util/SumoCommand.java
```diff
@@ -130,7 +130,7 @@ public SumoCommand(Object input1, Object input2, Object input3, Object[] array,
 
                 cmd.content().writeUnsignedByte(Constants.REQUEST_DRIVINGDIST);
 
-            } else if ((Integer) input1 == Constants.CMD_GET_SIM_VARIABLE && (Integer) input2 == Constants.POSITION_CONVERSION) {
+            } else if (input1 == Constants.CMD_GET_SIM_VARIABLE && input2 == Constants.POSITION_CONVERSION) {
 
                 if ((Integer)output_type == Constants.POSITION_ROADMAP) {
```
behrisch, line 135: You missed one, can you please add this one too?

behrisch: Thanks a lot, it is still much duplicated code but at least the casts are gone.

mernst: @behrisch Thanks for pointing out this oversight.  I probably missed it because it didn't have a space between the cast type and the expression.
You can take another look now.

behrisch:
Thanks again!

PR #7346, opened 2020-07-29
added multiprocessing capability to routeSampler.py

mschrader15:
We are frequently working with >= 24 hour data that we aggregate into 120 second intervals. Because of the large number of intervals, routeSampler is the longest running process in our workflow. I added multiprocessing capability to the routeSampler and drastically reduced its compute time.

I haven't tested with all combinations of option inputs, so this probably isn't "release" ready. It works for our purposes though so thought that I would share.   

As it stands, multiprocessing will not work with the mismatch-output option. 

Signed-off-by: Maxwell Schrader <mcschrader@crimson.ua.edu>

### DIFF
--- a/tools/routeSampler.py
+++ b/tools/routeSampler.py
@@ -471,7 +471,7 @@ def main(options):
         sumolib.writeXMLHeader(outf, "$Id$", "routes")  # noqa
         if options.cpuNum > 1:
             # call the multiprocessing function
-            results = multi_process(options.cpuNum, intervals, solveIntervalMP, outf, options=options, routes=routes,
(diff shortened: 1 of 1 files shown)


namdre:
Thanks Max!

namdre:
@behrisch should we rename the option **--cpu-num** to **--threads** (because it is more consistent with existing options) or do you think this is too misleading since python doesn't do threads.

behrisch:
> 
> 
> @behrisch should we rename the option **--cpu-num** to **--threads** (because it is more consistent with existing options) or do you think this is too misleading since python doesn't do threads.

Yes, **--threads** would be better. Doing it with "multiprocessing" is just an implementation detail ;-)

PR #7348, opened 2020-07-31
[Fix] spell of routing

finn0:
Fix spell of routing.


behrisch:
Thanks for the PR, we could not apply it due to missing paperwork, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work. But nevermind, I fixed it myself.

Issue #5353, opened 2019-03-22
Car-following model-based semantics of stop offsets


m-kro:
Related to #1285 and #3754 . Stop offsets advance the stop position of the vehicles upstream the stop line. The current implementation enables to configure the stop position locally (lane-based) and per vehicle class. A use case for this choice is given in #3754 : Cyclists can queue at the very front with motorised traffic behind.
What about considering the offset also from a driving perspective? Currently, this is captured for car following  a leader only but could be applied for stop lines, too. There is already the minGap attribute which represents this. Another factor*minGap offset could be introduced for stopping in front of an obstacle = stop line. Is there any technical reason against this (source code of MSVehicle mentions #1096 and #999)?

namdre:
I my opinion, nothing goes against introducing another car-following variable to configure stopline-gap.
Since stopOffset and endOffset actually move the stopLine (from the viewpoint of the junction model), the new variable would be added on top of that. with some extra checks negative values might be possible as well to allow for aggressive stopping.

Here are some questions:
- are stopLines the only relevant obstacle here?
- is there some interaction with the junction type? (in the current code, different stopping offsets are used depending on whether the link is major, minor or tls-controlled). See MSVehicle.cpp:2382-2402

PR #7535, opened 2020-09-10
make stop line gap configurable on vType base (#5353)

m-kro:
Instead of adding an additional variable as suggested by @namdre in #5353, the stop line gap replaces the constant DIST_TO_STOPLINE_EXPECT_PRIORITY.


namdre:
Thanks. We may still need to change this in the future because the stopOffset feature is meant to work like a new stop line and with the current implementation you cannot define a distance *relative* to the new stop line since its combined with `max`.

PR #7590, opened 2020-09-24
3 scenarios from Dublin added to the list

lcodeca:


lcodeca:
As always, when I try to change the wiki, the sign-off is broken.
Would you ming copy-paste the 3 additional lines, or should I try to revert and sign it again?

angelobanse:
👍

PR #7793, opened 2020-11-16
logging removed due to improper use

lcodeca:
The [wrong] way I used the logging was impeding the use of proper logging in other pipelines that were importing the script.

Signed-off-by: Lara CODECA <lara.codeca@gmail.com>


behrisch:
Thanks!

PR #7943, opened 2020-12-10
Can sumo count with Wireshark? What should I do?

KangChou:
Can sumo count with Wireshark?
What should I do?


namdre:
assuming this is a duplicate of #7944

PR #7989, opened 2020-12-16
Compile Java classes for libsumojni.jar with target version 1.8

kschrab:
The `libsumojni.jar` should be compatible with as many Java versions as possible. Currently this is not the case, as the target version of the compiled classes are not defined, making it dependent on the compiler used during the build. As a result, previous versions of libsumojni.jar had different major versions: in 1.7.0 it was build with a Java 8 compiler, in 1.8.0 with an Java 11 compiler, and in the nightly master builds currently uses a Java 15 compiler. However, you could simply define the target version during compiling the classes by using `-target` parameter.

Signed-off-by: Karl Schrab <karl.schrab@fokus.fraunhofer.de>


behrisch:
I don't know Java well enough but would reading https://stackoverflow.com/questions/54843571/java-11-what-is-the-difference-between-javac-release-and-source-and-target/54843883 it seems to be better to use `--release` instead of `-target`. What do you think?

kschrab:
Good point, I did not know about this option yet. This flag is available since Java 9, but as I remember your builds require a Java 11 compiler at minimum for LiSuM, right? If this is the case, I'll fix this to use `--release 8` instead.

kschrab:
hm, apparently some of your linux builds still use JDK 8

kschrab:
something went wrong during updating my fork, I'll close this PR

PR #8015, opened 2020-12-18
Compile Java classes for libsumojni.jar and libtracijni.jar with target version 1.7

kschrab:
The libsumojni.jar should be compatible with as many Java versions as possible. Currently this is not the case, as the target version of the compiled classes are not defined, making it dependent on the compiler used during the build. As a result, previous versions of libsumojni.jar had different major versions: in 1.7.0 it was build with a Java 8 compiler, in 1.8.0 with an Java 11 compiler, and in the nightly master builds currently uses a Java 15 compiler. However, one could simply define the target version during compiling the classes by using `-target` parameter.

We should not use the new parameter `--release`, as this is only available since JDK 9 but some linux builds still use JDK8. See discussion in #7989 . 

This PR replaces #7989 which has been messed up by myself 🤦🏻 

Signed-off-by: Karl Schrab <karl.schrab@fokus.fraunhofer.de>


behrisch:
I just saw that Traas even targets 1.7 with -source and -target Should we do the same here just for the sake of consistency? And could you add it for libtraci as well? Thank you!

kschrab:
Done, some tests are failing but I guess they're not related to my changes as they seem to fail on master too

behrisch:
Thanks a lot

Issue #7347, opened 2020-07-30
additional duarouter arguments for randomTrips.py


apiszcz:
Would it be possible to a parameter that would pass through arguments for duarouter? in particular the number of processing threads option?

namdre:
You can do this with a two-step process:
1. let randomTrips generate trips
2. call duarouter on those trips

JCodyA:
Hello!  This might be a good one for me but I'm pretty new to this and will need some guidance.  I was looking through randomTrips.py a little and trying to figure it out.  Will this enhancement be as simple as adding a parameter to a constructor or something like that?

behrisch:
Hi, this should add a generic way to provide more arguments to the underlying duarouter call.  It is a little bit more than just adding arguments to constructor. It should allow calls like `randomTrips.py duarouter--duaoption` to pass `--duaoption to the duarouter much like it is done in https://github.com/eclipse/sumo/blob/master/tools/assign/duaIterate.py#L464. Maybe the best way is to move the function https://github.com/eclipse/sumo/blob/master/tools/assign/duaIterate.py#L386 to sumolib and just reuse it.

JCodyA:
Ok this is making more sense.  Where should I move the function to, options.py or do I need to create a new file/class?

behrisch:
Yes options.py looks good to me. Thanks for picking it up!

JCodyA:
Sure thing!  I'll start working on it.  I haven't contributed to sumo before but I have with other eclipse projects.  Do you use the [WIP] tag on pull requests to mark them as works in progress?  Normally I would open a [WIP] PR quickly to allow comment and get help, which will likely be needed.

behrisch:
We haven't used this tag yet but it seems to be a good idea. Just go ahead.

namdre:
@bcoueraud87 for you

behrisch:
If you pick this up, please also have a look at https://github.com/eclipse/sumo/pull/8127

namdre:
randomTrip writes the used options into the header of the generated trips.xml. This doesn't play nicely with options of the form `duarouter--some-option`.

Solution 1: escape the double dashes when writing the options header (ugly)
Solution 2: Instead of re-using the ugly suffix-double-dash notation from duaIterate.py, use a prefix that looks more like a "normal" option:     

use prefix `--duraouter-some-option` and simply replace the `--durarouter-` prefix with `--` before passing the option to the duarouter application.

[...]

PR #8127, opened 2021-01-25
[WIP] Move assign_remaining_args function to allow reuse

JCodyA:
Move function to options.py to allow general reuse as mentioned in issue #7347 

Closes:  #7347 
Signed-off-by: J. Cody Arnholt <j.cody.tbone@gmail.com>

### DIFF
--- a/tools/assign/duaIterate.py
+++ b/tools/assign/duaIterate.py
@@ -23,8 +23,7 @@
 Run duarouter and sumo alternating to perform a dynamic user assignment.
 Based on the Perl script dua_iterate.pl.
 """
-from __future__ import print_function
-from __future__ import absolute_import
+from __future__ import print_function, absolute_import
 import os
 import sys
 import subprocess
--- a/tools/assign/duaIterate.py
+++ b/tools/assign/duaIterate.py
@@ -36,7 +36,7 @@ from costMemory import CostMemory
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
 import sumolib  # noqa
-from sumolib.options import get_long_option_names  # noqa
+from sumolib.options import get_long_option_names, assign_remaining_args  # noqa
 
 
 def addGenericOptions(argParser):
--- a/tools/assign/duaIterate.py
+++ b/tools/assign/duaIterate.py
@@ -23,7 +23,8 @@
 Run duarouter and sumo alternating to perform a dynamic user assignment.
 Based on the Perl script dua_iterate.pl.
 """
-from __future__ import print_function, absolute_import
+from __future__ import print_function
+from __future__ import absolute_import
 import os
 import sys
 import subprocess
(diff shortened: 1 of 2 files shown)


JCodyA:
This is a first step and I opened this pr now to allow guidance and comment.

behrisch:
Now you need to call the function from its new location in duaIterate.py to retain the old behavior

JCodyA:
So should I just add ```from sumolib.options import assign_remaining_args``` to the import section of duaIterate.py?  Sorry I just wanted to check first.  I thought the ```import sumolib``` line covered it.  This is a learning experience for me as well to be honest.

behrisch:
You can just add it to the existing line https://github.com/eclipse/sumo/blob/44f333bcc8f234b73de99b1cfc1e547ef926b293/tools/assign/duaIterate.py#L39 with a comma.

JCodyA:
Sounds good.  I'll do that, and the same with https://github.com/eclipse/sumo/blob/44f333bcc8f234b73de99b1cfc1e547ef926b293/tools/assign/duaIterate.py#L26-L27
so that it becomes ```from __future__ import print_function, absolute_import```  If everything checks out I can squash my commits into one if that is the preferred method here.

behrisch:
Please don't mix in unrelated refactoring changes.

behrisch:
I just noticed that get_long_option_names is just used in the moved function so there is no need to import it.

JCodyA:
Looks like everything is passing now.  I can go ahead and squash everything into one commit if you like.  Is there a tag or anything I should put in the commit comment body like [skip ci] so that it doesn't do all the tests again and waste processing?  Thanks for your help by the way, this has been a learning experience for me.

behrisch:
Well, this was only the first step, now we need to actually use that function in randomTrips.py.

JCodyA:
I'm looking through randomTrips.py now and trying to figure it out.  I feel that I have a learning curve here but I will keep working.  Any guidance is definitely appreciated.

behrisch:
The next step is probably to assign the remaining args in randomTrips.get_options to options.remaining_args and then pass them to the function you just moved. (The best way would be probably to refactor randomTrips to use the new argParser instead of the old OptionParser but that is maybe too much for now.)

JCodyA:
Ok, this is making more sense once again.  Would you like me to try and do the refactor?

behrisch:
Only if you are familiar with optParse and argParse and the differences. Otherwise just assign the parsed args to options.remaining_args.

behrisch:
@JCodyA Are you still working on this?

JCodyA:
I haven't since we last connected here.  Life has intervened a little.  I'm happy to continue but I still have to prioritize my time

behrisch:
Thanks for getting back so fast. I don't want to add stress just wanted to check whether you need help.

JCodyA:
Thank you!  No stress added whatsoever, and I will need help if I remember right but I'll need to look over everything again and refresh myself.

Issue #3477, opened 2017-10-03
track person state changes just like vehicles (trac #3477)


behrisch:
traci.simulation.getDepartedPersonNumber
traci.simulation.getDepartedPersonIDList
...

Migrated from http://sumo.dlr.de/ticket/3477
```json
{
    "status": "new", 
    "changetime": "2017-10-03T19:42:07Z", 
    "description": "", 
    "reporter": "behrisch", 
    "cc": "", 
    "resolution": "", 
    "_ts": "1507059727548557", 
    "component": "TraCI", 
    "summary": "track person state changes just like vehicles", 
    "priority": "major", 
    "keywords": "", 
    "time": "2017-10-03T19:42:07Z", 
    "milestone": "1.0.0", 
    "owner": "", 
    "type": "enhancement"
}
```

namdre:
this has been requested repeatedly via mailing list

Domsall:
For a simulation with Carla, I am looking into a way to not only transfer the vehicles from SUMO to Carla, but also the persons. Therefore I need to check when they departed and arrived. Have you done anything in this regard yet? If not, I can create a possible solution.
I looked into the source code for the vehicle arrived and departed functions and found the VehicleState and VehicleStateListener Class. I was now wondering if its smarter to add something like the VehicleState (e.g. TransportableState) in the MSNet class or to keep it under the MSTransportableControl class?! I think it depends if it may somehow be needed in another function later on.

namdre:
We haven't worked on this yet. I would put the TransportableStateListener code into MSNet to highlight the commonalities with the VehicleStateListener.

PR #8289, opened 2021-02-26
Adding TransportableState and getDepartedPerson and getArrivedPerson commands

Domsall:
originiates from #3477. I created the TransportableState just like the current VehicleState with the same commands. Other transportable state changes or API-commands haven't been added, but should be possible to easily add them now.

I tested the commands with a python skript and it worked. I didnt check other ways though. Maybe you can add some kind of test or have the time to check it yourself.

### DIFF
--- a/src/libsumo/Helper.cpp
+++ b/src/libsumo/Helper.cpp
@@ -114,6 +114,7 @@ std::vector<Subscription> Helper::mySubscriptions;
 Subscription* Helper::myLastContextSubscription = nullptr;
 std::map<int, std::shared_ptr<VariableWrapper> > Helper::myWrapper;
 Helper::VehicleStateListener Helper::myVehicleStateListener;
+Helper::TransportableStateListener Helper::myTransportableStateListener;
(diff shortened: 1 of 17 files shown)


namdre:
with this patch, 2 tests in tests/sumo/pedestrian_model/striping/errors are crashing

Domsall:
thank you very much for the check. Sorry for the incomplete pull request, I could have also done the standard test run myself... I found the problem, but yes need a bit of help. It seems that I haven't found the right spot to call the TransportableState. I found different variables and functions: myNumActivePedestrians, myRunningNumber and ::add, ::remove, etc..
Do you know which call or variable actually results in creating and destroying a transportable object?! Or a Person Object?! It's pretty clear, that I didn't quite understand the different levels (Person.cpp, MSPModel_....cpp, MSTransportableControl.cpp, etc.). I wanted to use the Departed Variable when a person actually departed, independant of being in a vehicle or walking (one could check that with already existing traci commands, could also be added to the state later on).

namdre:
I found the problem and will fix it after merge.

Issue #6330, opened 2019-11-27
DriverState device only affects carFollowModel Krauss


namdre:
@schwamborn I think this impacts on TransAID

schwamborn:
@namdre Thanks for the heads up! All configs I checked use Krauss as manualType, so our results shouldn't be affected by this. Nevertheless, also pinging @yunpangfloetteroed @robertalms

robertalms:
@schwamborn What about about the ACC/CACC models we use in TransAID?

namdre:
They are not supposed to suffer from the human imperfections that the driverstate device models.

robertalms:
Ah, correct. Then our results should be fine.

robertalms:
As I understand it, the 'applyDriverState' is deactivated by default. So that means, that the addition shouldn't affect the ACC/CACC model behavior or the TransAID results at all. I guess, this issue could be closed then.

namdre:
agreed.

PR #8290, opened 2021-02-26
Updated CACC by adding perception errors

mehdi-maleki:


namdre:
@schwamborn concerning the discussion in #6330
- allowing driverstate for CACC seems to go against the model intention but there should be no technical reason against allowing this combination (might be construed as modelling sensor errors maybe)
- this PR would change the behavior of simulations that mix Krauss and CACC and set option **--device.driverstate.probability 1**. The inputs would have to be changed to specify driver state device by vType rather than with an option. Do you foresee a problem?

behrisch:
@mehdi-maleki Thanks for this one! Unfortunately the required paperwork seems to be missing. Do you have an Eclipse account and accepted the ECA, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work ?

schwamborn:
@namdre I generally agree with the first point. As you pointed out, however, it seems to go against the model's intention, which is why I would prefer driverstate to be opt-in and make the user switch it on explicitly (maybe separately via a CFModel parameter like `carFollowModel.caccApplyDriverState`).

mehdi-maleki:
Hi,

Thanks for the update!

Yes, I have an Eclipse account and I signed the ECA (you can see in the
picture below).

@jakob if you mean driver awareness, yes I agree with you (since the
assumption is there is no driver in ACC and CACC). But in order to avoid
this problem when activating the driverstate device in CACC model, we
defined the initial awareness as 0.99 for the target vehicle. In this case,
we are able to introduce errors with the "Error State" parameter, which
allows us to model the sensor errors in speed and gap estimation.

Best regards,
Mehdi

[image: image.png]

namdre:
driverstate is now applied for CACC model when activated by vType attribute `applyDriverState="1"`

mehdi-maleki:
Great, thanks!

PR #8326, opened 2021-03-05
Added slope to ParkingArea spaces

alexbrummer:
Added slope to ParkingArea spaces to prevent transmission of INVALID_DOUBLE_VALUE for the slope of parked vehicles via TraCI. This way, also parked cars have a valid slope value.
Stating the slope of spaces is optional, so users who work with pure 2D scenarios are unaffected by the changes. In this case, the slope value is simply 0.
Also integrates with netedit.

Signed-off-by: Alexander Brummer <alexander.brummer@fau.de>

### DIFF
--- a/data/xsd/additional_file.xsd
+++ b/data/xsd/additional_file.xsd
@@ -487,6 +487,7 @@
         <xsd:attribute name="width" type="positiveFloatType" use="optional"/>
         <xsd:attribute name="length" type="positiveFloatType" use="optional"/>
         <xsd:attribute name="angle" type="xsd:float" use="optional"/>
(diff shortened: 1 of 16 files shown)


namdre:
Thanks for the contribution!

When testing, I noticed that the slope value in fcd-output would jump for road-side parking vehicles
- the sign of the slope when parking didn't match the sign of the slope just before parking
- road side spots at the apex of a hill had a different value compared to the previous value. While driving, the slope is simply taken to be the slope of the vehicle front rather than by taking the angle between front and back. While your version is more realistic, I prefer to avoid the discontinuity.

PR #8350, opened 2021-03-12
Double waitBetweenRetries between failures of traci connection.

the0demiurge:
When using traci in a slow cpu, traci keeps failing when starting. This is because after making a subprocess of sumo, the sumo process starts really slow while traci waits for only 10 seconds.

This PR will double waitBetweenRetries for slow cpus.

### DIFF
--- a/tools/traci/main.py
+++ b/tools/traci/main.py
@@ -124,6 +124,7 @@ def connect(port=8813, numRetries=10, host="localhost", proc=None, waitBetweenRe
             if retry < numRetries + 1:
                 print(" Retrying in %s seconds" % waitBetweenRetries)
                 time.sleep(waitBetweenRetries)
+                waitBetweenRetries *= 2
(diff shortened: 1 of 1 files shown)


behrisch:
Why don't you simply increase numRetries?

the0demiurge:
Double wait time would be more reasonable while most users would like to choose the default arguments, and trying to connect every second is not efficient.

behrisch:
I don't think so. Getting a fast second try has some benefit, at least I see that most of my traci experiments succeed at the second connection attempt which would take one second longer with your change. Maybe we could increase the waiting time after the second attempt? Or simply raise the default for the number of tries. Do you really experience a further slowdown of your machine just because your script tries to connect every second?

the0demiurge:
Actually, trying every second work well, add trying times to 30s or 60s would be a great help:)

behrisch:
OK, we increased the default to 60 now

Issue #4485, opened 2018-08-29
Import sumo libraries in another cmake projects


alexampa:
## Import sumo in a cmake project
Hello! I'm trying to import sumo, in particular libsumo, into a cmake project using command find_package(libsumostatic) in the corresponding CMakeLists file. However, when the project is being built the output is as follows:

```
...
Could not find a package configuration file provided by "libsumostatic"
  with any of the following names:

    libsumostaticConfig.cmake
    libsumostatic-config.cmake

  Add the installation prefix of "libsumostatic" to CMAKE_PREFIX_PATH or set
  "libsumostatic_DIR" to a directory containing one of the above files.  If
  "libsumostatic" provides a separate development package or SDK, be sure it
  has been installed.
...
```
Is there a way I could achieve importing the libsumo interface into my project? If not, could this be added as a feature? Thanks in advance!

### Specifications

  - Ubuntu version: 16.04
  - SUMO version: 0.32.0
  - SUMO branch: master

namdre:
Please try with the latest development version instead of 0.32.0

alexampa:
I'm sorry, I made a mistake about the version. I actually cloned the latest one (in particular 2853b1d), from the github repository and I followed the build instructions in the README file and in SUMO wiki for building and installing the SUMO binaries on Linux using cmake. The commands I used in detail are the following:
```
git clone --recursive https://github.com/eclipse/sumo
export SUMO_HOME="/path/to/sumo"
cd sumo && python tools/build/version.py
make -f Makefile.cvs
./configure && make install
mkdir build/cmake-build && cd build/cmake-build
cmake ../..
make
make install
```

behrisch:
You are doing both build options (cmake and autools) at once, which is not necessary. You should have a working sumo already after the first make install. Or if you'd rather do the cmake build, the following should suffice:

    git clone --recursive https://github.com/eclipse/sumo
    export SUMO_HOME="/path/to/sumo"
    mkdir build/cmake-build && cd build/cmake-build
    cmake ../..
    make
    make install
Also usually you need to have root privileges to do make install (so you have to do `sudo make install`) but your error message does not look like this is the problem. Can you upload / post the complete build log somewhere?

alexampa:
@behrisch 
Thanks for the reply. I do have root privileges and I don't have any problem building the sumo project, everything is built fine.

[...]

PR #8356, opened 2021-03-13
Initial CMake export configuration (#4485)

aaron-bray:
Currently:
 - libsumocpp is the only project part of the export
 - A minimum of projects are installing headers

Please let me know which (all?) CMake targets under src you would like to be installing its headers
I only updated the util projects that impacted my current (very minimal) usage

When/If you wish to export static libs, simply replace `install_headers` with `export_target`

I also set the default install path to a subdirectory of the build
On windows, the default `C:\Programe FIles\...` always causes issues
So, only if it is NOT set by a user, will this change to something local and less error prone...
(I generally always use a projects installation, not its build, but that is my preference)

To ensure this is working:
Look at the install directory when the build/install is complete, you should now see
`build/install/include/sumo` : This directory will have the same hierarchy as the src folder, only with headers
`build/install/bin` : On windows, this will hold the libsumo.cpp
`build/install/lib` : This will hold the libsumocpp.lib (and libsumocpp.so on *nix)
`build/install/lib/cmake/SUMO` : Here is where the SUMO CMake config package will be placed

For users to use SUMO in their CMake project they can:
```
# Set this to the folder that contains the SUMOConfig.cmake file installed by SUMO, or in the SUMO build folder
set(SUMO_DIR "C:/builds/sumo-master/install/lib/cmake/SUMO")
# You can also use the build directory
set(SUMO_DIR "C:/builds/sumo-master")
# both should work, again, I prefer using an install, but I know plenty of folks who only use the build directory
find_package(SUMO) # Required or not is up to the user
# link a library or executable to SUMO
add_library(target ${files} SUMO::libsumocpp)
add_executable(target ${files} SUMO::libsumocpp)
```


behrisch:
I don't think we need to publish the headers for the utils. I agree that there is a library built but it is only used in one project and is not documented at all.

aaron-bray:
I am still a novice at using the SUMO API, but what I am doing is just running a simulation and grabbing some vehicle information as it runs. To do this I need to include 

```
#include <libsumo/Simulation.h>
#include <libsumo/Vehicle.h>
```
These files include files from `common`, `geom`, `vehicle`, and `xml` 
So in order for my cpp to compile, all these files needed to be part the export
That is why I had to add the install_headers macro into those CMake projects

Depending on how users use the C++ API, more CMake projects will need to export their headers for their code to properly find and compile their code using SUMO.

For example, if someone wanted to code some C++ in their project that used the routing utilities, the `src/utils/router` CMake would need to "export" those headers.

A CMake expot package should "contain" all the headers and libs needed. 
That package is intended to be used in place of the source code.
There are two types of exports that are simultaneously supported by CMake, a build export and an install export.

1. The build export is in your build directory. It does not copy the includes. The include paths it adds to the export are paths to the source code used in the build. This export is only good on your machine.
2. The install export is located where you set the CMAKE_INSTALL_PREFIX to, this copies up all the headers identified in the export as well as the libraries. This is a fully relocatable SDK. (like an apt dev package)

So I think its worth thinking about adding all projects headers to the export, and maybe a few projects not to include, like gui?

behrisch:
OK, I see your point but this is rather a bug. The .h files in libsumo should be the only header files needed (without the ones ending in Helper actually) to use libsumo. I just noticed that this is not the case: #8472.

aaron-bray:
Ok, yes, that would simplify things!

Hopefully that can be easily fixed by forward declaring classes where needed in headers and moving the includes to the cpp files

aaron-bray:
@behrisch 
I removed the utility project installs from my initial implementation
I rebased to the current master with the libsumo header changes
SO only libsumo is installing headers now, and my code compiles against this

Looking good!

HuaweiAlgolux:
> @behrisch I removed the utility project installs from my initial implementation I rebased to the current master with the libsumo header changes SO only libsumo is installing headers now, and my code compiles against this
> 
> Looking good!

Hi Aaron, thank you for this info. I'm a dummy for libtraci. How do define `find_package()` in your cmakelist.txt for libtraci? I used `find_package(sumo REQUIRED PATHS /usr/share/sumo) ` , because `SUMO_HOME` is  `/usr/share/sumo`, but this did not work.

aaron-bray:
@namdre I don't think this MR is OBE
This MR adds the CMake logic to create the sumoConfig.cmake file that is needed by external users like @HuaweiAlgolux 

I can try to rebase this MR up to the current master
I believe this is still needed, but I have not dug into any CMake changes in sumo since this MR was created...

HuaweiAlgolux:
> @namdre I don't think this MR is OBE This MR adds the CMake logic to create the sumoConfig.cmake file that is needed by external users like @HuaweiAlgolux
> 
> I can try to rebase this MR up to the current master I believe this is still needed, but I have not dug into any CMake changes in sumo since this MR was created...

First, Thanks a milliion for reply in a quick manner. I'm desperately trying to make my project work. I'm still super super confused about how to use libtraci to interact with sumo-gui in my project. I'm trying to write a C++ program that actively controls the speed of a vehicle in sumo-gui via libtraci. But I have no idea which library to add to my CMakeList.txt. I saw you were using something like this `add_executable(target ${files} SUMO::libsumocpp)` in your CMakeList.txt. But my question is why is not `add_executable(target ${files} SUMO::libsumo)`? why is not `add_executable(target ${files} SUMO)` . I have never seen libsumocpp in the documentation. All I see is that libtraci is part of libsumo, not libsumocpp. And libtraci is SUMO's c++ interface

HuaweiAlgolux:
> @namdre I don't think this MR is OBE This MR adds the CMake logic to create the sumoConfig.cmake file that is needed by external users like @HuaweiAlgolux
> 
> I can try to rebase this MR up to the current master I believe this is still needed, but I have not dug into any CMake changes in sumo since this MR was created...

Can you share me your` sumoConfig.cmake` file?

aaron-bray:
@HuaweiAlgolux I have rebased my forked branch with the latest master from today
You should be able to pull my fork and build it, and you will get a sumoConfig.cmake you can use

HuaweiAlgolux:
> @HuaweiAlgolux I have rebased my forked branch with the latest master from today You should be able to pull my fork and build it, and you will get a sumoConfig.cmake you can use

Wow. Thank you sooooooooooooooooo much Aaron. I can't thank you enough. Would you mind if I offer to buy you an Amazon or Starbucks gift card?

HuaweiAlgolux:
> @HuaweiAlgolux I have rebased my forked branch with the latest master from today You should be able to pull my fork and build it, and you will get a sumoConfig.cmake you can use

Is this the correct branch [CMakeExport#4485](https://github.com/aaron-bray/sumo/tree/CMakeExport%234485)?

aaron-bray:
Yeah, clone my fork and check out that branch

behrisch:
Hi Aaron, thanks for updating the patch. I had a look and it changed the installed filesystem layout so I tried to simplify it, please see https://github.com/eclipse/sumo/commit/27eba3d2340de495da8541241d17efc9bdbc9cd3 I would like to hear your feedback!

aaron-bray:
Hi @behrisch 
The libs are being installed properly
The headers are not, and we still need to work on the export package.
I am not sure if you want me to make a MR for this or not
I know CMake can be pretty fusterating to learn, apply and maintain
So maybe we just go over a few changes/additions via this conversation to impart a little CMake knowledge so you know whats going on and can easier maintain? (I am also happy to get on a video call if you want to talk through some CMake questions)

So, reguarding the includes
The libsumo includes are actually being installed via [this call](https://github.com/eclipse/sumo/blob/27eba3d2340de495da8541241d17efc9bdbc9cd3/src/CMakeLists.txt#L91)
This is a valid pattern, but its not installing associated with a target, just copying files
I suggest you delete this function call, and we associate header installs via the target

So to associate the header install to each target, we want to use the PUBLIC_HEADER install variable
Note , the install INCLUDES does not actually copy files. When used, you are actually specifying a path that will be appended to end users include paths when using your target, which is something we should do, but we still need to specify how the target copies headers to the proper location
 
So here is a good post on using the [install PUBLIC_HEADERS variable](https://stackoverflow.com/questions/10487256/cmake-how-to-properly-copy-static-librarys-header-file-into-usr-include), by defining a variable list that contains all the headers you want to install

So now the libsumo/CMakeList should like this

```
set(libsumostatic_PUBLIC_HEADERS
    Edge.h
    GUI.h
    Helper.h
    InductionLoop.h
    Junction.h
    Lane.h
    LaneArea.h
    MultiEntryExit.h
    Person.h
    POI.h
    Polygon.h
    Route.h
    Calibrator.h
    BusStop.h
    ParkingArea.h
    ChargingStation.h
    RouteProbe.h
    Rerouter.h
    VariableSpeedSign.h
    MeanData.h
    OverheadWire.h
    Simulation.h
    Subscription.h
    StorageHelper.h
    TraCIConstants.h
    TraCIDefs.h
    TrafficLight.h
    Vehicle.h
    VehicleType.h
)
set(libsumostatic_STAT_SRCS
    Edge.cpp
    Helper.cpp
    InductionLoop.cpp
    Junction.cpp
    Lane.cpp
    LaneArea.cpp
    MultiEntryExit.cpp
    Person.cpp
    POI.cpp
    Polygon.cpp
    Route.cpp
    Calibrator.cpp
    BusStop.cpp
    ParkingArea.cpp
    ChargingStation.cpp
    RouteProbe.cpp
    Rerouter.cpp
    VariableSpeedSign.cpp
    MeanData.cpp
    OverheadWire.cpp
    Simulation.cpp
    TrafficLight.cpp
    Vehicle.cpp
    VehicleType.cpp
)

add_library(libsumostatic STATIC ${libsumostatic_STAT_SRCS} ${libsumostatic_PUBLIC_HEADERS})
set_target_properties(libsumostatic PROPERTIES PUBLIC_HEADER ${libsumostatic_PUBLIC_HEADERS})
add_dependencies(libsumostatic generate-version-h)
set_property(TARGET libsumostatic PROPERTY FOLDER "libsumo")

set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_DEBUG ${CMAKE_SOURCE_DIR}/bin)
set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_RELEASE ${CMAKE_SOURCE_DIR}/bin)
add_library(libsumocpp SHARED ${libsumostatic_STAT_SRCS} ${libsumostatic_PUBLIC_HEADERS})
set_target_properties(libsumocpp PROPERTIES PUBLIC_HEADER "${libsumostatic_PUBLIC_HEADERS}")
set_property(TARGET libsumocpp PROPERTY FOLDER "libsumo")
set_target_properties(libsumocpp PROPERTIES PREFIX "")
set_target_properties(libsumocpp PROPERTIES OUTPUT_NAME libsumocpp${BINARY_SUFFIX})
set_target_properties(libsumocpp PROPERTIES OUTPUT_NAME_DEBUG libsumocpp${BINARY_SUFFIX}D)
install(TARGETS libsumocpp EXPORT SUMOConfig
    PUBLIC_HEADER DESTINATION include/sumo/libsumo
    LIBRARY DESTINATION lib
    INCLUDES include/sumo
)
```

Note I am installing the headers to a sumo directory under the shared include directory.
The install folder can be shared by multiple projects all copying their includes up
So you want to create a project based directory you can put your project folders under
This will play well with other projects

And end users can #include "sumo/libsumo/Vehicle.h"
It makes their code a little cleaner and isolates all your includes under a single directory

You need to implement the same install pattern for libtraci
So once completed, when you install, the include dir will now have
include/sumo/libsumo
include/sumo/libtraci

Does that make sense?

Once this is done, we can talk about creating export packages
Which is a little more complicated

behrisch:
Hi Aaron, thanks for the fast feedback and yes let's discuss this here and yes cmake is hard to learn and to maintain and I am very thankful for your help. The setup is a little different though: libsumo and libtraci use the same header files that's why I thought it might even be a good idea to separate the header installation from the library installation. What exactly is the problem with the current setup? The generated config was usable for my small test project (well I tested only the installed one, I am not even sure about the use case for the one in the build tree).

aaron-bray:
I assumed that you would want to install Connection.h and Domain.h out of
libtraci?

Not sure if end users need those files when using the api

behrisch:
No, they don't.

aaron-bray:
Ah ha, I see what you are doing with libsumo and libtraci
That is an interesting pattern, generally, just looking at a code base I make the assumption that a directory is a target and headers files in that directory are associated with only that target. 
I thought libtraci would be dependent on libsumo, but I see both libsumo and libtraci are implementing the same class definition differently. 
Again, I am just making assumptions of sumo's intentions based on the directory structure, which is pretty subjective. 
I would probably put the headers in their own directory rather than libsumo, but that is just reflective of how I interpret source directory structure

So I get the installation logic now
Associating headers to targets is generally better for projects that optionally export multiple targets, and target oriented CMake is the 'Modern CMake' design pattern that is pretty popular these days. But that too is subjective, and what you have is working fine for what sumo is doing (exporting 2 targets that share the same include set)

So moving on to the export configuration (SUMOConfig.cmake)
Notice you actually have 2 SUMOConfig.cmake files
One in your installed directory (install/share/sumo/SUMOConfig.cmake)
And a SUMOConfig.cmake in your build directory

The installed config looks good.
You can see that it defines targets for the libraries
and it also associates the include directory so projects automatically add that dir to their include path.

If you look at the SUMOConfig.cmake in your build directory
You will notice that it does not have any include directory information for the provided library targets

So what is the deal with 2 config files?
Projects do not install by default.
The install needs to be explicitly called, there needs to be an install location defied by CMAKE_INSTALL_PREFIX, but the install is a relocatable directory structure
As a user of SUMO, I may or may not run the install target
So I may not get the install config`
But I do always build the code, so I will always get a build dir SUMOConfig.cmake
That build dir config is generally what developers use when using find_package
they will set SUMO_DIR to the sumo build dir, then call find_package(SUMO)
And they will actually include files directly from the SUMO source code


If you open the SumoConfig.cmake file in the build dir, there is nothing in it that identifies the src/libsumo directory as an include dir

To get a good SumoConfig.cmake file in the build directory that adds sumo source directories to the include path,
add the target_include_directories line below to both the libsumo and libtraci CMakeLists.txt
```
add_library(libsumocpp SHARED ${libsumostatic_STAT_SRCS})
set_property(TARGET libsumocpp PROPERTY FOLDER "libsumo")
set_target_properties(libsumocpp PROPERTIES PREFIX "")
set_target_properties(libsumocpp PROPERTIES OUTPUT_NAME libsumocpp${BINARY_SUFFIX})
set_target_properties(libsumocpp PROPERTIES OUTPUT_NAME_DEBUG libsumocpp${BINARY_SUFFIX}D)
target_include_directories( libsumocpp
    PUBLIC
      $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/src>
)
install(TARGETS libsumocpp EXPORT SUMOConfig
    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
    INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
)
```

So you want to test that your example project that is using the provided sumo package works with SUMOConfig.cmake in both the install and build directory
```
set(SUMO_DIR <install/share/sumo/cmake>)
find_package(SUMO REQUIRED)
```

and

```
set(SUMO_DIR <sumo/build>)
find_package(SUMO REQUIRED)
```

You never know which users of SUMO will want to use...

Note that the Modern CMake pattern is to install your config file in the install/lib/cmake folder (i.e. install/lib/cmake/sumo), which just means that is where people are generally first looking for config files in a cmake install dir

So I think just adding that BUILD_INTERFACE location is really it, and maybe changing where you are putting your install config

behrisch:
Thanks a lot for the detailed explanation. I added the BUILD_INTERFACE as you recommended.

ghorzangasi (commented):
Leave it

Issue #8108, opened 2021-01-20
Public transport tutorial (without web wizard)


MPLoves:
Hello,

We want to study the subway trafic and more especially the waiting time of the passenger, and how to reduce it. 
Do do that we want to use SOMO and simulate a piece of subway, only 3-4 station on a straight line, but we really don't know how to start it.
Is there some exemple or a tutorial that we can follow please?

Thank you very much for the help

Best;

behrisch:
You can start with the WebWizard extracting your favourite subway line from OpenStreetMap https://sumo.dlr.de/docs/Tutorials/OSMWebWizard.html. Make sure to check the "Public Transport" checkbox. Unfortunately there is currently no public transport tutorial. I keep this ticket open as a reminder to write one.

lbieker:
ticket was closed in last commit

PR #8378, opened 2021-03-17
Public Transport Tutorial Markdown, ref #8108

LaraKaiser:
Signed-off-by: Lara Kaiser <Lara.Kaiser@Student.HTW-Berlin.de>

A tutorial for: Public transport tutorial (without web wizard) #8108
Describes the useage of buses and trams within sumo and guides the user through a small exercise to build the first simulation utilizing public transports.


namdre:
Some comments:
- add a link from the tutorials page (https://sumo.dlr.de/docs/Tutorials/index.html) to your new page 
- setting line="ANY" would create problems in bigger scenarios when different lines use the same stop. Thus it would be safer to set the line attribute of the trams and use that in the ride definition
- public transport typically drives according to a schedule (stop attribute 'until') and this is in fact a necessary condition for public transport routing (https://sumo.dlr.de/docs/Simulation/Public_Transport.html#public_transport_schedules) so I recommend adding this attribute as well in your tutorial

LaraKaiser:
Thank you for your feedback! 
I added the link and set the line attributes accordingly. I also wrote another 
chaper for the schedules, to make it a little clearer.

namdre:
lists require a leading empty line or they will not render correctly (e82deb4393f43b78fa24895d86bde26b9252ad3f)

angelobanse:
To force a new line, you have to put three spaces (instead of `\`) at the end of the line.

PR #8395, opened 2021-03-21
Update index.md

haneenHisham:
The 'SUMO_HOME' variable is not read until sourcing the .bash_profile file

### DIFF
--- a/docs/web/docs/Installing/index.md
+++ b/docs/web/docs/Installing/index.md
@@ -94,6 +94,10 @@ where `/your/path/to/sumo` is the path stated in the caveats section of the `bre
 ```
 echo $SUMO_HOME
 ```
+If the path is not printed you can try the following command
+```
+source ~/.bash_profile
+````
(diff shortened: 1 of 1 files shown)


namdre:
The documentation says to "Restart the Terminal" and this should source the profile.

behrisch:
> 
> 
> The documentation says to "Restart the Terminal" and this should source the profile.

No, I think restarting the terminal reads only .bashrc but not .bash_profile which is only read by login shells. So we should either recommend .bashrc (my preferred solution) or accept this change. But MacOS users may prefer different hings here @angelobanse ?

angelobanse:
I use `.zshrc` (since Z shell is the default since macOS Catalina) - where `source ~/.zshrc` and restarting the terminal does the same.

angelobanse:
We should recommend `.bashrc`/`.zshrc` and correct the following sentence on the macOS installation paragraph:
```
Depending on your shell, you may set this variable either in .bash_profile or .zshrc. To set this variable in .bash_profile you can use the following commands.
```

since `.bash_profile` and `.zshrc` are not equivalent.

PR #8502, opened 2021-04-16
Update FAQ.md

iamhyc:
fix a typo in section "Can SUMO be run in parallel (on multiple cores or computers)?"


namdre:
thanks (-:

PR #8520, opened 2021-04-20
macroscopic fundamental diagram code

Amirhoseinkr73:
Hello,
first of all, I am sorry if I send my request in the wrong file but this is my first pull request and I tried to do it in the right way but I don't know if this pull is requested in the right section or not.
I have developed a code that can aggregate data based on edge-based data and show fundamental diagrams ( density vs flow, density vs speed, and speed vs flow).

This code which is called MacrOutput describes the relationship between density, flow, and speed based on sumo Edge-based output. Edge-base output includes some information about density, speed, travel time and etc. to extract fundamental diagram based on Edge-based data, mean density, mean speed, and mean flow should be calculated. To calculate the mean of every component, data must be aggregated for all segments. In SUMO there is not any module to aggregate data and show fundamental diagrams. Also, aggregating data with some software like Excel is time-consuming. Therefore macrOutput is developed to cover this gap.  aggregate data and show macroscopic fundamental diagram (density vs flow, density vs speed, speed vs flow) based on simulation data. Also, MacrOutput demonstrates fundamental diagrams based on both density and lane density. Moreover, MacrOutput generates two output file which shows mean density, mean speed, mean flow results. The calculation process and relationships are taken from [1]. To use this feature execute the following steps:
1. Run the simulation and get the edge-base data file (for example “OUTPUT_FILE.xml”
2. Open OUTPUT_FILE.xml in Excel.
3. Add a column with the name “Length” and add the length of edges according to the Edge id column.
4. Save the file in CSV format (OUTPUT_FILE.csv) (usually it can be converted using Excel).
5. Put MacrOutput.py in the simulation folder.
6. Run cmd
7. Run this code (MacrOutput.py -i <input file (this case : OUTPUT_FILE.csv)>)

Reference
1. Qiong Lu, Tamás Tettamanti, Dániel Hörcher & István Varga (2019): The impact of autonomous vehicles on urban traffic network capacity: an experimental analysis by microscopic traffic simulation, Transportation Letters, DOI: 10.1080/19427867.2019.1662561.

Note : 
if the length of every Edge is showed in Edge-based data output in a column with the name "Lenght", users will not be forced to edit the file and add a column manually and users only must convert XML to CSV and run the code to get fundamental diagrams and results.

[MacrOutput.zip](https://github.com/eclipse/sumo/files/6349297/MacrOutput.zip)

Attach files:
1. this file includes the code and an example of a basic freeway network fundamental diagram


namdre:
I think a tool that plots the macroscopic fundamental diagram is a worthwhile addition.
However, the proposed way to use this does not match the usability standards of our tools.To be specific, the steps
- Open OUTPUT_FILE.xml in Excel.
- Add a column with the name “Length” and add the length of edges according to the Edge id column.
- Save the file in CSV format (OUTPUT_FILE.csv) 

are not acceptable.

If you need additional network properties you could instead open the network with sumolib and spare the user from these manual steps (https://sumo.dlr.de/docs/Tools/Sumolib.html#import_a_network_and_retrieve_nodes_and_edges)

However, in this case sumolib isn't even needed because edge lengths can be derived from edgeData output using the formula `length = sampledSeconds / period * 1000 / density`

Also, before doing your pull request you should place your tool in sumo/tools/visualization and commit it.  Before doing the pull request do a `pull --rebase` to make sure that only your commit shows up in the pull request rather than hunders of other commits. Feel free to do a new pull request from a fresh fork.

regards,
Jakob

Amirhoseinkr73:
Hello dear Jakob,
I have solved the problem of lengths and the user only must convert XML file CSV.  I mean the steps of running this code are as follows:
1. rum CMD
2. Run this code (MacrOutput.py -i <input file (this case : OUTPUT_FILE.csv)>)
I tried to develop a code that converts XML to CSV but it was not successful and I tried to use xml2csv.py but this code can not create columns in the right way. so I want to consult with you, is it okay to recommend the user convert it to CSV. because converting the edge-based output to CSV using Excel is very easy (just dragging file into Excel and save as CSV). what is your opinion about it? 
best regards,
Amir,

namdre:
xml2csv supports options for setting the separator character (i.e. colon or tab) and should be able to support your use case.
 I do not wish to include a tool that requires an intermediate excel-step.

Amirhoseinkr73:
I agree with you about the excel step but I tried XML2CSV but the output file is not a correct CSV file and all the results in every time interval are in a cell and I think xml2csv is not the solution and during this time I try to use some ways to convert XML to CSV in a right way but unfortunately they were not successful. so I don't know how we can solve this problem and I think Excel is the only way to solve it.

namdre:
The simplest solution would be to not convert the xml to csv at all but read int directly instead. 
Example code for building a pandas dataframe from xml input can be found at https://github.com/eclipse/sumo/blob/master/tools/output/scheduleStats.py
See also https://sumo.dlr.de/docs/Tools/Sumolib.html#parse_all_edges_in_a_edge_data_meandata_file

Amirhoseinkr73:
Dear Jakob, I have solved this problem and I have changed the code and the user only must run cmd and input XML file and get the results. so if you agree I delete this fork and build a new pull request with new code.

namdre:
That sounds great! Please do.

Amirhoseinkr73:
I have created a new pull request in visualization ....  please check it : https://github.com/Amirhoseinkr73/sumo/pull/1
thank you so much

Issue #8065, opened 2021-01-09
document LISUM version compatibility and add backward compatibility option


namdre:
version 1.9 is (currently) only compatible with LISA version 7.2+ (https://github.com/eclipse/sumo/commit/5dee0724d99683b7bdbcb9800a24a433deb1cd5f) whereas older version of LISUM are not.

Add switch to allow backward compatibility for older LISA version by selectively disabling the protocol change.

PR #8545, opened 2021-04-23
Lisum version compatibility

Philhil:
it is not the best an logic way to define a min Version. I am open for improvements   

ref #8065


behrisch:
Thanks for looking into it! (There was already a ticket #8065 targeting it.) I found your solution a little too heavy and added my own in https://github.com/eclipse/sumo/commit/555d0b75f9a184e839e692f920f52e7c2c5c93ce. Please recheck whether it fits your needs and open another PR if you want to fix/improve it.

PR #8563, opened 2021-04-28
Add license information to libsumo/libtraci pom.xml

kschrab:
There are maven plugins (e.g. `org.codehaus.mojo:license-maven-plugin`) which check the license of artifacts. This requires the pom to include a `<licenses>` tag.


namdre:
@behrisch We should also add the secondary license here, right?

behrisch:
yes, we should

behrisch:
@kschrab Thanks!

PR #8576, opened 2021-04-30
Added Person-TraCI commands to 'old' API, added net.netOffset and changed Person-SpeedFactor

Domsall:
I changed/added a few TraCI-commands and therefore created a pull request:
- Added a new command to `getParameter` to get the netOffset via TraCI-C++-Command
- Added the commands `getSpeedFactor`, `moveTo`, `moveToXY` and `setSpeedFactor` to the "old, deprecated" TraCI-API, as I will still need a bit of time to transfer to libtraci (I already have a dublicate libsumo-structure that works fine).
- Changed the `getSpeedFactor` and `setSpeedFactor` commands, because they didn't get or set the ChosenSpeedFactor, but rather get the value from the VType. Also setting the value didn't really transfer to the ChosenSpeedFactor correctly.

If everything looks good --> thanks for adding! If not, feel free to change it.

### DIFF
--- a/src/libsumo/Person.cpp
+++ b/src/libsumo/Person.cpp
@@ -363,7 +363,7 @@ Person::getLength(const std::string& personID) {
 
 double
 Person::getSpeedFactor(const std::string& personID) {
-    return getPerson(personID)->getVehicleType().getSpeedFactor().getParameter()[0];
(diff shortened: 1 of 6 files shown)


namdre:
Thanks!

PR #8592, opened 2021-05-04
Fix capitalization of W99 car following parameters

marecabo:
SUMO complains if the capitalization from the documentation is used and not the one from [SUMOXMLDefinitions.cpp#L722-L730](https://github.com/eclipse/sumo/blob/master/src/utils/xml/SUMOXMLDefinitions.cpp#L722-L730).


namdre:
Thanks for the notice! I didn't merge this PR directly because your ECA wasn't in order.

marecabo:
I know, I wanted to do this tonight and therefore converted the pr to a draft...

Issue #8020, opened 2020-12-18
parameterized subscriptions should include size of parameters


behrisch:
Currently the traci server needs to parse the parameters just to know the size. Furthermore we need to store which variables need parameters at all on the server side making the protocol unflexible. This is a protocol breaking change and should be discussed.

PR #8601, opened 2021-05-06
traci: don't try to read parameters for subscriptions that don't have them

sommer:
I just tried out SUMO 1.9.1 with Veins 5.1 and encountered a breaking bug: 74d2f044c899fc185ff47689bc0780676cb8ab37 introduced a more flexible way of reading parameters for object subscriptions, but both `VAR_FOLLOWER` (of `CMD_SUBSCRIBE_VEHICLE_VARIABLE`) and `VAR_TELEPORT_ENDING_VEHICLES_IDS` (of `CMD_SUBSCRIBE_SIM_VARIABLE`) have the same value (`0x78`). This leads `TraCIServer::addObjectVariableSubscription`  to expect parameters for `VAR_TELEPORT_ENDING_VEHICLES_IDS`, instantly breaking the connection.

I am attaching a crude patch that works around this by tracking "this requires a parameter" not just for variables, but for a combination of command-and-variable (using the value `0` as a wildcard)

### DIFF
--- a/src/traci-server/TraCIServer.cpp
+++ b/src/traci-server/TraCIServer.cpp
@@ -292,10 +292,10 @@ TraCIServer::TraCIServer(const SUMOTime begin, const int port, const int numClie
     myExecutors[libsumo::CMD_GET_OVERHEADWIRE_VARIABLE] = &TraCIServerAPI_OverheadWire::processGet;
     myExecutors[libsumo::CMD_SET_OVERHEADWIRE_VARIABLE] = &TraCIServerAPI_OverheadWire::processSet;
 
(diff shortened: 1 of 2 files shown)


behrisch:
@sommer @namdre I'd rather do #8020 unless you think this breaks too much existing code.

sommer:
As far as the next release of SUMO is concerned I have no strong feelings either way. We just increase the API version (because, I’m guessing, the API now expects different parameters for the same commands) and we’re done: SUMO version X works after everybody updates their TraCI clients. Might be worth releasing a bugfix for the 1.9 series though. Otherwise we’re left with the 1.9 series never having worked with TraCI clients who subscribe to teleporting vehicles.

Issue #8232, opened 2021-02-15
optionally sort GTFS output


behrisch:
The route file is currently unsorted which is fine for most applications which load it as an additoinal anyway but not for all.

LaraKaiser:
Currently the vehicles are sorted by their ids within the route file. Would the ticket be completed by adding an additional option to sort them by their departure time?

behrisch:
Well yes and no. The option itself does not do the sorting yet but it should be not too hard to invoke sort_routes

PR #8724, opened 2021-06-07
added sort GTFS output #8232

LaraKaiser:
The filter_trips method now calls sort_routes on the created route file.

Signed-off-by: LaraKaiser <Lara.Kaiser@Student.HTW-Berlin.de>


behrisch:
Thanks for this one and sorry for the delay. Could you please recheck whether this also works with python2 and make the sorting optional?

LaraKaiser:
Thank you for the feedback! I checked the code with Python2.7 and it worked fine. The user can now choose if the sorting should be included with the --sort option.

behrisch:
Thanks! What is `os.path.join(os.getcwd(), ...` for? Looks like it should work without as well.

behrisch:
Never mind I just merged it and we will do the cleanup afterwards

PR #8750, opened 2021-06-10
traci: added loadState saveState

guetlein:
Signed-off-by: Moritz Gütlein <mrtzz@mail.de>


namdre:
Thanks for the contribution. Please note that utils/traci is deprecated and will not be actively updated by the core team.
The recommended successors are libsumo and libtraci which provide a consistent interface for TCP and directly library use of TraCI from C++

PR #8764, opened 2021-06-15
Ev powertrain device 1 8 0

kevbad:
Hello everyone,
 
my colleagues and I developed an EV energy consumption model for a paper titled
"Accurate physics-based modeling of electric vehicle energy consumption in the
SUMO traffic microsimulator" which we submitted to the IEEE ITSC 2021 in April.
It has been integrated into SUMO as a device named "EV Powertrain Device" and
explicitly models the individual components of the powertrain instead of
relying on a single constant efficiency for propulsion and recuperation,
respectively. For example, the power loss in the electric motor is interpolated
from a two-dimensional characteristic map and the battery power is based on an
internal resistance model.

Our contribution comes with five XML files that define electric vehicles from
different segments which are equipped with the EV Powertrain Device. All
parameters were taken from publicly available sources like manufacturer spec
sheets. For the most part, the device utilises the same XML attributes as
SUMO's Battery Device, introducing only a handful of new ones for the EM's
power loss map, the gear ratio, etc. This means that the devices are (mostly)
compatible with one another, though direct comparisons have shown that our
model is more accurate. Results are logged using a dedicated output class whose
XMLs can easily be converted into CSV files with an included Python script.

We would like to give back to the SUMO community by adding our model to the
official repository. If you're interested, let me know about any changes you'd
like to see prior to pulling. Please note that we must keep our fork as we've
referenced the branch "ev_powertrain_device" in the paper above.


namdre:
awesome!

behrisch:
Thank you very much for this contribution. So let me ask first: how much more time are you willing to invest here? I think we will definitely integrate it but we need to move things into different places. I am just pondering whether we refactor things in this PR (where it is probably easier for you to contribute even more) or if we merge first and refactor later (which makes it easier for us).

kevbad:
What exactly needs to be done? Putting a couple of hours into refactoring is no problem.

behrisch:
> 
> 
> What exactly needs to be done? Putting a couple of hours into refactoring is no problem.

We should move everything to the SUMO code style. I would start with directory structure (there is no real need for a toplevel vka directory and the name is not very descriptive ;-)) My first impression was that the constants and helper functions also duplicate code which is already in sumo. I would try to remove those two and move the evpowertrain maybe to utils/emissions. In general it might be worthwhile to rethink whether the powertrain should rather be used as an emission model and not as a device. What do you think?

kevbad:
We had a premonition that you'd want to move our files into the existing folder structure but didn't do it ourselves lest they end up in the wrong places. :-)

Using the powertrain as an emission model does sound like the better alternative. If I'm not mistaken, that would also integrate it neatly into the GUI and the XML outputs that are already there. However, it looks like a more complicated endeavour to me as our powertrain relies on a `CharacteristicMap` object to model the electric motor and the `PollutantsInterface`'s `compute()`/`computeAll` method only supports `double`s as additional parameters.

behrisch:
I think we could have a separate model for every CharacteristicMap muc like in the PHEM model. Maybe we should have a phone call to discuss this.

kevbad:
In the case of our EV powertrain, characteristic maps determine the power loss in electric motors as a function of speed and torque. Hardcoding them into a handful of emission models means that users will either be limited to the vehicle definitions we've provided or have to manually create new models if they want to simulate the power consumption *accurately*. It is possible to use the map of one vehicle in another (on condition that their EMs are comparable) but the loss of accuracy one would suffer kind of defeats the purpose of utilising our model. Also, the `CharacteristicMap` class does not extrapolate. If the demanded operating point of the motor lies outside of its map, the power loss is set to zero and the state flagged as invalid.


> Maybe we should have a phone call to discuss this.

That sounds like a good idea. I'll send you an e-mail with my telephone number.

kevbad:
Alright, time for a status update. Michael and I had a phone call at the beginning of July and came to the conclusion that our contribution should be integrated as a SUMO emission model, similar to PHEMlight. So far, I've implemented most of the cosmetic changes that were asked for, i.e. using SUMO's built-in helper functions and removing our namespace. The remaining ones, such as adopting the project's comment style for file banners and the Doxygen documentation, will follow soon.


While working on the realisation of the emission model, I came across a few issues where your knowledge and opinions would be highly appreciated:
1. All emission models get a base index `b << 16`. The lower 16 bits are used to identify emission classes, i.e. different configurations of the model in question, with the most significant bit being reserved for marking heavy vehicles. Did I get that right?
2. Is it standard procedure to register emission classes during runtime through calls to `PollutantsInterface::Helper::getClassByName()`? Is this where I should instantiate new objects of `SimpleEvPowertrain`?
3. I'd like to keep on using [our XML files](https://github.com/mechatronics-RWTH/sumo/tree/ev_powertrain_device_1_8_0/data/VehicleParams) in the emission model. What's the easiest way to parse them inside a `PollutantsInterface::Helper` in order to access the `vType`-parameters?

Best regards,
Kevin

namdre:
re 3: the result of my discussion with @behrisch is that 
- we'll change the interface of PollutantsInterface::Helper::compute to receive an instance of Parameterised instead of a `std::map<int, double>`. 
- the batteryDevice will pass an appropriate instance populate from the vType and vehicle params (so you do not need custom parsing code)
- the Parameterised instance will support the method  `std::vector<double> getDoubles(key, default)`  which may use caching in the background for faster access to complex values. (likewise caching may be added to the existing method Parameterised::getDouble) (possibly even with an `int key` overload and more caching)

namdre:
Refactoring (#9254) is finished.
To use your own attributes you should do the following:
- add the strings and enum values for all your attributes to src/utils/SUMOXMLDefinitions.h / .cpp
- add defaults for doubles and lists of doubles here: https://github.com/eclipse/sumo/blob/12fb4e3887d59f7056db570b8f8a2f8001c25662/src/utils/emissions/EnergyParams.cpp#L32-L48
  - initialization from vType `<params>` as used in your xml file then happens automatically
- within your new PollutantsInterface::Helper, use params.getDouble, params.getDoubles

kevbad:
The power loss map string isn't merely a list of floating point numbers, it contains information about the dimensions of the map, its axes, and entries. The `CharacteristicMap` class doesn't know how to interpret the data without the different kinds of delimiters. You can find an in-depth explanation of the string format and the class [here](https://github.com/mechatronics-RWTH/sumo/blob/ev_powertrain_device_1_8_0/src/utils/emissions/MMPEVEM/CharacteristicMap.cpp).

behrisch:
While I agree that this is probably the more readable and maintainable solution, I don't think the computational overhead would be so big if we reduce the characteristic map to a set of functions operating on three vectors of doubles. The internal storage is a vector of doubles anyway and the intensive part is the interpolation as far as I can see.

kevbad:
How would you like to proceed?

namdre:
So far, sumo is only loading xml files and I'd like it to stay that way.
Putting all the energy related attributes/maps etc into multiple `<param>` elements of the `<vType>` (i.e. with `key="powerLossMapString" value="2,1|0,413.7931,827.5862...."/>` would be my preferred solution.

All the code for loading and manipulating these maps can go into utils/emissions/EnergyParams. This will ensure that data is loaded exactly once and then kept around for efficient usage.

Please let me know if I missed something. I'm happy to give more support and otherwise await your pull request.

kevbad:
I've implemented your preferred solution, @namdre.

namdre:
Great! The new code doesn't break any test, looks good style-wise and we'll be happy to merge it. We may even be able to include it for next weeks release of 1.11.0 

- A minor change is needed for MSFrame.cpp (new options where added in the meantime causing a merge conflict). 
  - I'd prefer the options to be named `ev-powertrain-output` instead of `ev_powertrain-output`
  - there isn't currently any code that makes use of the new option to write powertrain output (please either remove the options and add the code later or add more code)
- do you have an example scenario that uses the new model?
- do you have a paper reference that I can put in the documentation
- do you have a piece of example input for convertMapMat2XML.M ?

kevbad:
* > there isn't currently any code that makes use of the new option to write powertrain output (please either remove the options and add the code later or add more code)

  The `ev_powertrain-output` option was a remnant of the original implementation and has been deleted.



* > do you have an example scenario that uses the new model?

  You can take one of the `vType` definitions in *data/emissions/MMPEVEM* and simply generate vehicles of that type.



* > do you have a paper reference that I can put in the documentation

  Our paper is titled *Accurate physics-based modeling of electric vehicle energy consumption in the SUMO traffic microsimulator* (DOI: 10.1109/ITSC48978.2021.9564463).



* > do you have a piece of example input for convertMapMat2XML.M ?

  The script is used to convert the output of the [electric machine design tool](https://github.com/TUMFTM/Electric_Machine_Design) developed by Kalt et. al. (DOI: 10.1109/EVER.2019.8813601) to our format.



Let me know if you need more information.

namdre:
@kevbad It would be great if you could flesh out the documentation at bit more. I've only added a stub at 
https://sumo.dlr.de/docs/Models/MMPEVEM.html

Issue #8835, opened 2021-07-16
tools/createVehTypeDistribution.py to sumolib


mschrader15:
Hey All,

Any interest in porting over [createVehTypeDistribution](https://github.com/eclipse/sumo/blob/master/tools/createVehTypeDistribution.py) to sumolib? My use case is running many simulations with varying bounds on car-following parameter distributions in a Monte-Carlo analysis. The workflow currently has to include a `subprocess.run()` call, as well as writing a temporary text file.  It'd be nice if I could use sumolib instead.

I'd be happy to adapt createVehTypeDistribution to fit the sumolib convention and create a pull request, but I wanted to first check and see if it was something that the SUMO team sees as beneficial.

behrisch:
Yes that would be great!

mschrader15:
Quick question: which versions of Python does sumolib aim to support? 

Wondering if I can use https://docs.python.org/3/library/dataclasses.html or not?

behrisch:
> Quick question: which versions of Python does sumolib aim to support?

In general we still try to support 2.7 if possible but the main platform is 3.7 and above.

> Wondering if I can use https://docs.python.org/3/library/dataclasses.html or not?

If it is not too much hassle please don't

mschrader15:
Ok, makes sense. I avoided using it in https://github.com/eclipse/sumo/pull/8844

PR #8844, opened 2021-07-19
Implementing https://github.com/eclipse/sumo/issues/8835#issue-946597310

mschrader15:
This is my first attempt at implementing #8835.

Let me know if any version issues / code structure changes desired

Example usage is commented out at the bottom of tools/sumolib/vehicle/vehtype_distribution.py


behrisch:
I think it is a little overengineered. You could just add the code to some sumolib/vehicletype.py. Also I think we should refactor the old script to use the new code, but we can also do it in a second step<

namdre:
@mschrader15 can you do the refactoring of createVehtypeDistribution.py?

behrisch:
@mschrader15 I just noticed during [refactoring](https://github.com/eclipse-sumo/sumo/commit/5db4e80fa11af0686042e091bd9b5cf403651d10) that there is no test for your CreateMultiVehTypeDistributions. Any chance you can contribute a small realistic test case here? If not, I will make up something.

PR #8861, opened 2021-07-27
Fix libsumo headers

riebl:
After installing a SUMO CMake build with `make install`, I cannot use the installed libsumo headers in a third-party project without these modifications.


behrisch:
Thanks a lot!

Issue #8492, opened 2021-04-15
Allow distribution string for more vehicle type parameters


marecabo:
Currently, only the [speedFactor](https://sumo.dlr.de/docs/Definition_of_Vehicles%2C_Vehicle_Types%2C_and_Routes.html#defining_a_normal_distribution_for_vehicle_speeds) of a vtype allows to be filled by a distribution string e.g. `"normc(1,0.1,0.2,2)"`. Although there exists a [script ](https://sumo.dlr.de/docs/Tools/Misc.html#createvehtypedistributionspy)for generating vehicle types with distributed parameters, it would be a great simplification, if the distribution string could be used with for example `tau` or other numeric vehicle type parameters.

behrisch:
This would probably result in a separate vType being instantiated for every vehicle much like it is now done when vType parameters are modified via traci. This makes it harder to do a "batch" modification of vehicles by changing the associated vType but this probably a rather rare use case.

namdre:
We already have a way to model inheritance among vehicle types (getOriginalID) and use this when filtering by vtype id (detectors, calibrators and rerouters). This might be used (optionally) for batch modifcations via TraCI as well.

PR #8909, opened 2021-08-04
Adding Extended IDM to master

Domsall:


namdre:
Thanks for the contribution!

namdre:
I'd like to see a publication reference in the EIDM comment header. Shall I use "Extending the Intelligent Driver Model in SUMO and Verifying theDrive OffTrajectories with Aerial Measurements"?

Domsall:
Yes thats the one. Feel free to change Comments/Code to your coding standards

namdre:
More questions:
- can you document the new model and its parameters at 
https://sumo.dlr.de/docs/Definition_of_Vehicles,_Vehicle_Types,_and_Routes.html#car-following_models
https://sumo.dlr.de/docs/Definition_of_Vehicles,_Vehicle_Types,_and_Routes.html#car-following_model_parameters

- the drive_in_circles tests show several collisions when running at step length 1. What is the recommended maximum step length for EIDM?  (Our IDM implemenation performs an internal iterates to compute safe speed at higher step length)

- the stop_at_position tests show that EIDM stops ahead of a specified stop position. While this can be solved on the user side by always defining a stop startPos with sufficient slack, it may lead to surprising results when swapping models.  Do you think it's possible to fix this?

- it seems that each EIDM vehicle clones it's type in order to randomize it's minGap. I think this would be betters solved via #8492 (eventually).

Domsall:
- I'll do the documentation, how can I change the site?
- Yes, this is kind of expected. I first used the internal iteration, but had problems with the way I used some variables and therefore had to delete it again. The model works good from 0.05-0.5s. The idea was to have a good model for subsecond and submicroscopic simulation (I work for an automotive institute and use SUMO together with Carla, Simulink, etc. and need realistic acceleration patterns). I normally use 0.1s.
- I am going to take a look at that tomorrow, I know that back then I had a lot of problems with the min_gap and positions when stopping. Are the values/differences rather big or small? The internal "human perception errors" are expected to cause imperfect stopping.
- I have two different models (EIDM and DEIDM) and probably forgot to take that out again for the EIDM, I will follow with another commit

namdre:
- documentation is in the main repo and you can simply do pull requests against https://github.com/eclipse/sumo/blob/master/docs/web/docs/Definition_of_Vehicles%2C_Vehicle_Types%2C_and_Routes.md
- the tests show a stop position ~0.8m ahead of the intended target. The default slack for stopping on a lane with only 'endPos' defined is 0.2m

namdre:
@Domsall There are still some unused variables in the code that should either be removed or handled with the UNUSED_PARAMETER macro  and kept for later. Which one would you prefer?

Domsall:
feel free to change it to your UNUSED_PARAMETER macro. I will work on these additions in the future:

- adding vehicle dynamics to the model, not only for emission/energy calculation, but also for acceleration thresholds at different speeds and for different vehicle types (mass, etc.)
- adding spatial anticipation to reduce collisions and create "traffic theory" waves.

Domsall:
I normally compile the SUMO Code with VS 2017 (on windows). Warning Level 4 is already turned on and I added /WX to get errors when unused parameters are found, but I could not find any...

Did you already change the Code?

namdre:
Yes. @behrisch already fixed this in 50a4d27f1351cb730755883b5611f5b1af2a9d29

Issue #8565, opened 2021-04-28
gracefully handle xml comments in parse_fast and parse_fast_nested


namdre:
- if there are comments after the first data lines have been found, issue a warning (might be a multi-line comment)
- if a comment starts on a line, don't parse anything after the comment start
- header comment should be ignored (no warning)

LaraKaiser:
I would like to take a look at it.

PR #9096, opened 2021-09-07
Modified file for issue #8565

LaraKaiser:
Added a method to prefilter the file for comments and throw warnings if necessary, to interfere as little as possible with the original code and file.
Tested for Python 2.7 and 3. 
Signed-off-by: LaraKaiser <Lara.Kaiser@Student.HTW-Berlin.de>

### DIFF
--- a/tools/sumolib/xml.py
+++ b/tools/sumolib/xml.py
@@ -343,14 +343,31 @@ def parse_fast(xmlfile, element_name, attrnames, warn=False, optional=False, enc
     @Example: parse_fast('plain.edg.xml', 'edge', ['id', 'speed'])
     """
     Record, reprog = _createRecordAndPattern(element_name, attrnames, warn, optional)
-    for line in _open(xmlfile, encoding):
+    _tempfile_xml(xmlfile)
+    for line in _open("temp.xml", encoding):
         m = reprog.search(line)
         if m:
             if optional:
                 yield Record(**m.groupdict())
             else:
                 yield Record(*m.groups())
+    os.remove("temp.xml")
 
+def _tempfile_xml(xmlfile):
+    """
+    Creates new xml file 'temp.xml' with the input data stripped of all the comments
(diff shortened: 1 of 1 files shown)


namdre:
Can you quantify the impact on processing speed? (i.e. on 1GB of fcd-output)  If creating a stripped copy of the original input is much slower we may want to make this optional.

LaraKaiser:
Yes, sure. I'll take a look at it!

LaraKaiser:
I tested the _tempfile_xml with several file-sizes and got the following results:
1,13 GB:
12.5202865s

805 MB:
8.8093499s

685 MB:
7.4966952s

260 MB:
2.9469349s

namdre:
Can you put this in context with the before/after-pull-request running time when parsing a file?

LaraKaiser:
Yes, it does make quite an impact on the processing speed:

1.3 MB:
2.9000000000278803e-06

685 MB:
1.2999999999818712e-06

260 MB:
9.000000000258801e-07

With added code: 1.3 MB:
12.975254099999999

With added code: 685 MB:
7.6413338

With added code: 260 MB:
0.0011410000000005027

namdre:
I see three options
- merge as is (downside: make the operation much slower for people that often do not need this but who actually rely on the speed of parse_fast)
- make the comment filtering optional (downside: people unaware of the comment problem may receive invalid results / errors)
- find a faster implementation

any opinions @behrisch ?

behrisch:
I think it should be possible to do this in one go. Essentially you only need to modify the regex to check for comment starts as well. Ia gree it makes the code uglier but it should keep the parsing speed.

behrisch:
Thank you. The current approach does not work if the comment is only a part of the line. Is that right?

behrisch:
Thanks! I reworked it a little and since this PR had merge conflicts I applied it separately in ca1a3ab0ac6976ebadb9b1c265355c6b76b09649

PR #9179, opened 2021-09-17
Improved platoonmanager

schwoz:
The improvements are these:

The former feature reduces the number of unnecessary catchups and the whole simulation becomes quieter and less chaotic. I have not yet performance figure though.

The latter prevents vehicles to be pushed on a lane with dead end (or occupied) short before the intersection. This measure prevents vehicles getting stuck, at least for the default lanechange model. With the sublane model it can still happen that vehicles get stuck in roundabouts. Increasing the strategy factor and pushy factor helps to improve this situation, but does not resolve it completely.

Parameters are set in __init__():
        self._edgenumberLookahead = 3

-  Or the ego vehicle needs at least this distance of common route length with leader before agreeing to follow.
        self._distLookahead = 500.0

- no lane change advice if vehicle has less than this distance to the next  junction

Touched only functions:
- _updateVehicleStates(self):
- _adviseLanes(self):

### DIFF
--- a/tools/simpla/_platoonmanager.py
+++ b/tools/simpla/_platoonmanager.py
@@ -1,5 +1,5 @@
 # Eclipse SUMO, Simulation of Urban MObility; see https://eclipse.org/sumo
-# Copyright (C) 2017-2021 German Aerospace Center (DLR) and others.
+# Copyright (C) 2017-2020 German Aerospace Center (DLR) and others.
 # This program and the accompanying materials are made available under the
 # terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


namdre:
@schwoz I'm still seeing this check fail: "The author(s) of the pull request is not covered by necessary legal agreements".
Did you already perform the necessary eclipse-related steps (https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work) ?

schwoz:
I have now done the recommended steps for eclipse, I close now and  I'll retry to do a pull request from patch-1

schwoz:
OK deleted patch-1 do pull request from my fork

PR #9213, opened 2021-09-22
Fast, dedicated EmissionsOutput to CSV Parser

mschrader15:
We are working with large [emission output files](https://sumo.dlr.de/docs/Simulation/Output/EmissionOutput.html) (> 1-2 Gb). 

This proposed `EmissionsXML2CSV` class drastically reduces (by ~95% compared to `tools/xml/xml2csv` with 1.5 Gb file) the time to convert the XML output to a more friendly CSV

I am creating a pull request only to propose the method and see if the SUMO team has interest in integrating. If so, I will clean up the code and perhaps move it somewhere else in the `sumolib` package.


behrisch:
Thanks a lot for this one! Since your approach seems to be generalizable would you like to make a general script out of this? Also did you try to compare your speedups with sumolib.xml.parse_fast?

mschrader15:
I hadn't seen `sumolib.xml.parse_fast` until last night when I added my code below.  If I would have found it yesterday morning, I probably wouldn't have written my own

Here is a timing comparison with a 1.5 Gb File:
- Proposed Method:
    - 8 cores
        - 8.6 seconds
    - 24 cores
         - 5.8 seconds
     - Optimization to do on chunk size vs number of cpu cores used 
- `sumolib.xml.parse_fast`:
   -  1 core 
       - 41.5 seconds
   -  [Here is how I implemented it](https://gist.github.com/mschrader15/2487f012a99a7f09087bb7265d8329e0), there may be a better way?  
- `tools/xml/xml2csv.py`
    -  1 core
        - 238.2 seconds

I really would be interested in extending to other use cases. I will get started on that...

behrisch:
That's still an impressive speedup!

mschrader15:
I'm thinking that I could implement my method by simply adding an optional `threads` argument to `sumolib.xml.parse_fast`? If the number of threads are specified, `EmissionsXML2CSV` (or a more generally named `ParallelXML2CSV`) handles the processing?

namdre:
- parse_fast returns a stream of python objects (and it would be very cool to make this faster via parallelization). For parse_fast, the objects must be flat (i.e. no timestep/vehicle) structure
- parse_fast_nested is for 2-level hierarchies (i.e. interval/edge or timestep/vehicle) and could handle emission output files
- xml2csv could have a switch for using parse_fast/parse_fast_nested if the user knows the depth of the object graph. Then you could have another switch for using a parallel version of parse_fast / parse_fast_nested

This would be awesome because all tools that use parse_fast could potentially make use of your parallelization speedup and it should be even faster than your current implementation.

mschrader15:
Duh. Completely forgot about the fact that parse_fast & parse_fast_nested return python objects. The slowest part of my parallelization is the file write thread, so simply returning a stream of python objects should cut down processing time by a second or more. 

Your layout make sense @namdre . I take a first pass at integrating parallelization into parse_fast and parse_fast_nested this weekend

mschrader15:
I tried to incorporate `ParallelXML2CSV` into `parse_fast_nested` by using the compiled regex expressions and made a couple realizations:

1. `namedtuple` does not play nicely with python multiprocessing. I tried the monkey patches highlighted [here](https://stackoverflow.com/questions/16377215/how-to-pickle-a-namedtuple-instance-correctly), but none worked (on Windows) for me
             - I attempted to replace `namedtuple` with [`SimpleNamespace`](https://docs.python.org/3/library/types.html#types.SimpleNamespace), which can be extended to capture most of the functionality of namedtuple, besides immutability, however using `SimpleNamespace` with mp.Queue is quite slow (aka slower than just calling `parse_fast_nested` alone.  
             - An alternative is to only send the regex groups to the main thread and let it handle the creation of the `namedtuple` instances. This doesn't lead to any speed up in my testing. The parallelization is ultimately only as fast as the Queue consumer. 
             
2. If the threaded XML parse function composes comma-separated strings and the main thread directly writes these strings to a file, then the speed up from my proposed code is preserved
         - To me, this means that the parallelization is only useful in the XML2CSV tool?

namdre:
Could you check whether a hypothetical function parse_fast_parallel that returns normal python tuples, preserves the speedup from parallelization? Having this as a library function (used by xml2csv.py) might still permit other tools to be made faster.

mschrader15:
Interestingly, the slowest part of the "tuple method" or even the SimpleNamespace appears to be the Queue.put()

With a list of "row" strings (~170000 comma separated strings) the queue.put() time is ~0.1 seconds. Putting that same data in the Queue in the form of 170000 tuples (or list) takes ~4-5 seconds. I can't explain this.

Using the row strings, I can split them back up in the main thread to form tuples, and the speed up is largely preserved.

namdre:
Hi Max, I just tried your patch and this is what I'm getting
- `from multiprocessing import shared_memory` isn't used but creates a dependency for all users of sumolib
- `yield from` breaks python2 compatibility. There should be a two-liner to achieve it's use with the simple old `yield`
- our tests for tools/sumolib/parse_nested are giving a different result than before

mschrader15:
Hey Jakob,

Sorry, I should have added a comment after my commit that **I did not intend for this to be merged,** only for you to review the code and see how I implemented the multiprocessing aspect.

I wasn't sure if you still wanted to incorporate the MP given my messages above. But now that I know you do have the intention of merging, I will work on the Python 2 compatibility and test on varying XMLs.

I am guessing that one of the differences between the tests is that the MP parser returns "chunks" out of order. I will implement a sorting algorithm to return the chunks in the file order.

mschrader15:
I am closing this for now as I tried a couple of ways to fix
> our tests for tools/sumolib/parse_nested are giving a different result than before

and all solutions that I found led to a negligible speedup.

PR #9236, opened 2021-09-24
Improved Simple platoon manager2

schwoz:
**Improvement of platoonmanager**

Improvements are these:

Parameters are set on init:
        self._edgenumberLookahead = 3

- Or the ego vehicle needs at least this distance of commom route length with leader before agreeing to follow.
        self._distLookahead = 500.0

- no lane change advice if vehicle has less than this distance  to the next  junction


schwoz:
Sorry I could not figure out how to be covered by the legal agreements. I looked at my eclipse account and I am a contributer but not a committer, which should be fine. I just agreed to the most recent license, don't know if that helps.

namdre:
@behrisch any ideas what could be wrong here? Was the "signed-by" requirement or is it still needed?

behrisch:
No, maybe wrong associated email or the github account is not present in the Eclipse account settings?

behrisch:
@schwoz I also think this patch should be redone against the head of current main. Your fork seems to be severely out of date, that's why we have merge conflicts.

schwoz:
My github email is identical to the eclipse email (I also did the git config --global user.email "email@example.com")
OK I have now merged the current sumo main with my master and resolved the conflict. Maybe it works now.

PR #9252, opened 2021-09-27
 Improved Simple platoon manager3 (closed the old one)

schwoz:
**Improvement of platoonmanager**

Improvements are these:

Parameters are set on init:
        self._edgenumberLookahead = 3

- Or the ego vehicle needs at least this distance of commom route length with leader before agreeing to follow.
        self._distLookahead = 500.0

- no lane change advice if vehicle has less than this distance  to the next  junction


namdre:
@behrisch I'm unable to get more information on the failed eca validation ("This validation request is closed or unavailable.") Can you ask our friends at eclipse for support with this?

Issue #9163, opened 2021-09-15
support public transport priority for actuated traffic lights


namdre:

namdre:
alternative solution with custom switching rules is documented at https://sumo.dlr.de/docs/Simulation/Traffic_Lights.html#bus_prioritization

PR #9287, opened 2021-10-01
Support for Public Transport Priority (see #9163)

eriktunsch:


behrisch:
Is there a reason for deleting the AUTHORS file?

eriktunsch:
I wanted to remove the AUTHORS file from the pull request, because it was shown to me, that it would trigger a conflict.
I think, I missinterpreted the button "remove file"....

behrisch:
we want to do this differently now

PR #9542, opened 2021-11-12
Further update of device.elechybrid code

jprk:
This PR represents updates related to solvability of the overhead electric circuit, and further small improvements to the original `device.elechybrid` code. It also addresses the power vs. energy errors that have been already corrected in the electric energy consumption model.


namdre:
@jprk  We still have only 3 tests in our sumo.extra test suite for elecHybrid. Maybe you have more tests cases that you could share? Or do those three test capture all relevant code paths including outputs?

behrisch:
And please rethink whether the options should really be boolean with a default value of "true". While I appreciate avoiding double negation (something like "--no-recuperation false" is worse) it is often hard for users to understand how to set this option correctly. So maybe you can find another positive name? For the moment I also renamed them to fit our general scheme of separating the domain and the sub-option with a "."

namdre:
see a5f49caeb77db9f06141f14cf60770c64539c27c for the new option names.

jprk:
You are correct. I will clear it with @kubasevca, he probably has some
tests that did not make it into the test suite.
-- jan

jprk:
Second ACK. Thanks, @behrisch, for the code cleanup on weekend, by the
way, and for 58e4f3b today. I did not expect @namdre would be so fast
merging it in, and probably did not follow all the code style rules. I am
also afraid I accidentally packed two files that represent traction
substation stubs for the next update, I have to check that as well.

behrisch:
@jprk Can you please also review 3790132e0a0f262f0fd04c5d0b727022dfbbccb8. I hope it did not break anything

jprk:
@namdre @behrisch We have almost finished another set of updates as per your comments (tests, option modification). Shall I open another PR or do you prefer some other way to incorporate them into the current Sumo tree?

namdre:
fresh PRs please. One for tests and one for code changes.

PR #9568, opened 2021-11-17
Updated NEMAController getCurrentPhaseIndex 

mschrader15:
I wanted the SavedTLSStates XML output to contain a useful phase attribute for validation purposes. 

When using the NEMA controller, the phase output is now the combination of Ring 1 Phase * 10 + Ring 2 Phase


namdre:
Unfortunately, this doesn't work. Check the test sumo/basic/tls/NEMA/basic
Sumo assumes that the phase index changes, every time that even a single signal state states. Your updated phase index doesn't do this and it causes invalid switching in the simulation (even though the sequence of states looks ok in the SavedTLSStatesXML.
It may also cause further problems down the line when trying to access the traffic light via TraCI because the list of phases is not defined at the new index values (i.e. 44).

The NEMA controller sets the phase name (i.e. 4+4 or 2+6) and an easy fix for your requirement would be to include the phase name in the tls-xml outputs.

mschrader15:
Thanks for the explanation @namdre. I guess I need to learn how to run tests before creating pull requests...

I considered including the phase name in the tls-xml outputs, but thought that changing NEMA code would be more likely to be accepted. I will retry by adding ` << "\" phaseName=\"" << myLogics.getActive()->getCurrentPhaseDef().getName()` to **src/microsim/actions/Command_SaveTLSState.cpp**

namdre:
the output should be conditional on having a name != ""

PR #9575, opened 2021-11-17
Fix of Updated NEMAController getCurrentPhaseIndex #9568 

mschrader15:


namdre:
Thanks for the update. I noticed that the existing code was quite ugly and your code duplication didn't make it any prettier. So here's the functionality with some refactoring.

Issue #9848, opened 2021-12-23
Adding fixed force off mode to NEMA type


qichaow:
Adding fixed force off mode to NEMA type. Under this mode, non-coordinated phases can use unused time of previous phases.

PR #9880, opened 2022-01-04
add fixed force mode to NEMA; start the cycle based on offset

qichaow:
#9848


namdre:
hi @qichaow, I was expecting the resuts of the fixForceOff=true tests to change (they have better performance now). However, the behavior of the fixForceOff=false tests changed as well. If this is intentional it would be good to describe the semantics of the intended change (i.e. as a fixed bug) for our change log.

namdre:
The changes can be seen in 81dfa3c2ff0c1890310f3f7c5335601218665442. The previous test results did already reflect the updated phase timings, so any changes are due to the new code. It mostly looks like (small) performance improvements.

Issue #9894, opened 2022-01-06
I packaged SUMO as a Flatpak and would like to submit it to Flathub.


kevinsmia1939:
Hello,

I successfully packaged SUMO as Flatpak for Linux distros.

In summary, Flatpak can build and distribute applications independently of what is installed on the host distro. So I do not have to make a separate package for other distros. It usually consumes more storage space than necessary as it bundles the required dependencies with it. 

The Flatpak manifest (which contains instructions to build SUMO) is here: https://github.com/kevinsmia1939/flathub/tree/org.eclipse.SUMO

I gave it the name "org.eclipse.SUMO" to conform to the reverse DNS naming scheme, but if you have another name in mind, let me know.
https://docs.flatpak.org/en/latest/conventions.html#application-ids

I will also upload it to Flathub, a software store for Flatpak. 
https://flathub.org/home
https://github.com/flathub

I have not built SUMO with Java yet, it complains that no javah executable is found. It does not look like javah exists in openjdk 17.

The package still missing metainfo.xml, I will create and submit it.

kevinsmia1939:
SUMO tried to download extra packages when it found maven.
But flatpak-builder does not allow network connection when building package.
How can I disable this? Or pre-download it.
I set -DCHECK_OPTIONAL_LIBS=false but still try to download.
```
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-jar-plugin/2.3.1/maven-jar-plugin-2.3.1.pom
[WARNING] Failed to retrieve plugin descriptor for org.apache.maven.plugins:maven-jar-plugin:2.3.1: Plugin org.apache.maven.plugins:maven-jar-plugin:2.3.1 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-jar-plugin:jar:2.3.1
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-source-plugin/3.2.0/maven-source-plugin-3.2.0.pom
[WARNING] Failed to retrieve plugin descriptor for org.apache.maven.plugins:maven-source-plugin:3.2.0: Plugin org.apache.maven.plugins:maven-source-plugin:3.2.0 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-source-plugin:jar:3.2.0
[...]

PR #9896, opened 2022-01-06
Add metainfo.xml for to briefly describe the software

kevinsmia1939:
Add metainfo.xml for to briefly describe the software. Which is useful for software store like Flathub for example.
This is relate to this. https://github.com/eclipse/sumo/issues/9894


behrisch:
Thanks for the patch! Could you please check whether you have an Eclipse account and have acknowledged the ECA, see https://github.com/eclipse/sumo/blob/main/CONTRIBUTING.md#eclipse-contributor-agreement

behrisch:
@kevinsmia1939 We would love to include this in the next release. Can you manage to make the paperwork by the end of the week?

kevinsmia1939:
not sure what I did, but the anonymous does not go away.

behrisch:
OK, it seems this one get messed up but I am sufficently convinced that you did the paperwork but tthis got unmergeable now. Should I simply use the contents of this file: https://github.com/flathub/org.eclipse.sumo/blob/master/org.eclipse.sumo.metainfo.xml (which is more recent) and place it in  build/package/sumo.metainfo.xml ?

kevinsmia1939:
> OK, it seems this PR got messed up but I am sufficently convinced that you did the paperwork. Should I simply use the contents of this file: https://github.com/flathub/org.eclipse.sumo/blob/master/org.eclipse.sumo.metainfo.xml (which is more recent) and place it in build/package/sumo.metainfo.xml ?

Yes, that is fine.

behrisch:
OK, did that, thanks for the PR. So now after the next release we can update the flathub repo and remove the metainfo file there, correct?

kevinsmia1939:
Yes we can, we can write a small command in the flatpak manifest to install the metainfo if it does not do it automatically.

PR #9953, opened 2022-01-15
Addressing dangerous barrier cross in #9940

mschrader15:
Opening this PR for #9940 . I would like @qichaow 's thoughts before merging, 

This is my first substantial C++ PR, so my feelings will not be hurt if major refactoring or changes need to made 😆

PR #9969, opened 2022-01-17
Implements green transfer according to #9954

### LINKED ISSUE #9954: NEMA should support Green Transfer
Expected behavior of Green transfer is (for example): 

1. In [3, 7]
2. Pass minimum timer on [3, 7]
3. Detector call on [3]. [7] goes into Grn XFER
4. Detector call on [8]. [7] enters transition immediately, [3] stays until extension timer exhausted

Current behavior is:

1. In [3, 7]
2. Pass minimum timer on [3, 7]
3. Detector call on [3]. [7] moves to [8] immediately
4. [3] stays until extension timer exhausted, then goes to [4]

The implementation will have to be outside of the nextPhase logic, as the timing will need to be adjusted. 

This is very similar to Rest in Green in #9937

_Originally posted by @mschrader15 in https://github.com/eclipse/sumo/issues/9937#issuecomment-1013740306_

mschrader15:
Fixes #9954.

This PR depends on #9953 and #9966. I will handle the rebase if/when those changes get implemented.


mschrader15:
This fails existing tests slightly at the beginning (around 10 seconds into simulation). There is a slight difference in initialization.

mschrader15:
Closing because logic is baked into #9970

Issue #9965, opened 2022-01-17
NEMA checks extension timer during yellow, which spills into next service duration


mschrader15:
During testing, I have caught an erroneous edge case @qichaow . The fix is easy, but it will cause differences in every single NEMA test. 

Below is an explanation of what I am seeing: 

The logic in https://github.com/eclipse/sumo/blob/fb134ce889e7543f54defcf6e9fa4fe6e678f0eb/src/microsim/traffic_lights/NEMAController.cpp#L759-L772 does not check whether the current phase is in Green or Yellow. During certain situations, this can cause the vehicle extension timer to be extend while the phase is transitioning to Red. 

Because the `phaseExpectedDuration` is only set to `0` on the [falling edge of Green](https://github.com/eclipse/sumo/blob/fb134ce889e7543f54defcf6e9fa4fe6e678f0eb/src/microsim/traffic_lights/NEMAController.cpp#L831), the `phaseExpectedDuration` can be overriden to be the duration + `vehExt` in https://github.com/eclipse/sumo/blob/fb134ce889e7543f54defcf6e9fa4fe6e678f0eb/src/microsim/traffic_lights/NEMAController.cpp#L761 when the phase is in yellow. 

Using @namdre 's improved phase tracker, we can see the behavior below. This comes from running `basic\tls\NEMA\fixForceOff\ramp_low_ffO_false`:

---

_We are looking at Phase 4+4 below_

**The current behavior:**

![image](https://user-images.githubusercontent.com/57813331/149791086-b46ca933-bc5c-461e-87d8-748e51f73b37.png)

Detector 05_0 actuates phase 4. We can see that a car arrives as phase 4 goes yellow at 3340. This cause the `phaseExpectedDuration` to no longer be 0 but instead the ` durationR1 + vehExt[R1Index]`. This `phaseExpectedDuration` spills over into the next time that 4 is served at 3445. 

---  

**The desired behavior:**

![image](https://user-images.githubusercontent.com/57813331/149791035-1de25b98-4753-4ac4-a302-35d7e170e8ed.png)

The extension timer should not be extended once the phase goes to yellow. This then resets the  `phaseExpectedDuration` to the minimum timer the next time that it is served.

mschrader15:
@qichaow Rainy and cold here in Alabama so not much else to do! Sounds good.

mschrader15:
#9970 closes this

Issue #9940, opened 2022-01-14
NEMA crossing ring barrier dangerously (serves crossing traffic together)


mschrader15:
@qichaow I was doing testing on a 4-way intersection and caught an undesirable barrier crossing.

The command line output looks like:

```
current time in cycle:  43.00   phases: 2       6
phase 3 was skipped
phase 4 was skipped
phase 1 was skipped
phase 7 was skipped
current time in cycle:  44.00   phases: 2       8
current time in cycle:  45.00   phases: 2       8
current time in cycle:  46.00   phases: 2       8
```

It appears to get caught in this error state and then never exiting. I have created a reproduce-able test case here: https://github.com/UnivOfAlabama-BittleResearchGroup/sumo/tree/4way-error/tests/sumo/basic/tls/NEMA/errors/4way_error

It maybe that I have some failing case of parameters, but I think the issue lies in the fact that `NEMALogic::nextPhase` can go all the way around the ring on the top ring `2->3->4->1->2` and in the same call to `NEMALogic::NEMA_Control`, R2 goes from `6 -> 7 -> 8`.

I have some ideas to fix this, but figured that I would hold off until you investigated @qichaow  

Let me know if you have questions & we can discuss further in the meeting tomorrow AM

namdre:
@mschrader15 Thanks for the test case! I've taken the liberty to putting it directly into main.

mschrader15:
No problem!

qichaow:
@mschrader15 I can see this happening with the minRecall function. I usually imagine minRecall on 2 &6 but the scenario you mentioned could happen. We should check the other ring and not cross barrier when do next phase. If you want to work on it, go ahead and do it. I can review it in your fork if you want. I can cover that if you don't have the time to do it now.

I appreciate your enthusiasm in this work. Keep it up.

mschrader15:
@qichaow I have implemented my solution on https://github.com/UnivOfAlabama-BittleResearchGroup/sumo/tree/4way-fix with commit. I also created a pull request #9953, with the purpose of making the changes more visible.

The main addition that I have added to the code is https://github.com/UnivOfAlabama-BittleResearchGroup/sumo/blob/101f6d62752dbab804bcdcd13da5410cea3887dc/src/microsim/traffic_lights/NEMAController.cpp#L1036 , which makes a decision on the next phase with knowledge of whether the other ring is ready to also change state and which "barrier" the other ring is on. 

[...]

PR #9970, opened 2022-01-18
Fix of #9965 & #9940

mschrader15:
@qichaow and I met to discuss the changes I proposed and decided to keep the changes that I proposed in #9953 & #9966 with minor refactoring. I closed those PR's in favor of a cleaner branch with less merges.

[`ff9e9f`](https://github.com/eclipse/sumo/commit/ff9e9f3af765319b07c2887f17e2937359075f71) updates the tests to match the results from fixing #9965 @namdre . I'm not sure the proper etiquette on changing test results so I can roll back that commit if you would like.

Another small change was implemented here: https://github.com/eclipse/sumo/commit/1cc8ec969f6ae7b84802b434e64d173ec5b81fe9#r64062447


qichaow:
Looks good. @namdre, I don't have issues with merging this pull request unless you have other concerns.

namdre:
Even though you did patch test results in this PR, these are the failing tests after merging:
Screenshot from 2022-01-18 07-19-38

namdre:
Some questions then:
- do the new results reflect a work-in-progress state that is expected to revert to the original results with the next PR, or do they reflect sensible expected behavior? (I seem to recall that you mentioned, some results would change back and forth in one issue but couldn't find it anymore)
- I assume that you didn't see the complex traci tests but do you have an idea why the 'fully_actuated' test is showing a different tls_state result?

On a side note, the test 4way_error was showing collisions, that were due to lack of internal lanes (coarser junction model) that doesn't play nicely with right-on-red. I've rebuilt the network and the collisions are now gone.

PR #9986, opened 2022-01-18
Enable high dpi awareness for the GUI applications

### LINKED ISSUE #9985: Enable dpi awareness for the GUI applications
Enable dpi awareness for Windows for netedit and sumo-gui.

Enabling the high dpi awareness flag in Visual Studio would make at least the netedit gui application no longer look blurry on higher than 100% dpi for Windows.

alexpagri:
Fixes #9985 

**Modifies CMakeLists**


behrisch:
If I read the docs correctly this requires a quite recent CMake. Can you please guard this with a cmake version check?

alexpagri:
https://cmake.org/cmake/help/latest/prop_tgt/VS_DPI_AWARE.html

_New in version 3.16._

However I now see that "cmake_minimum_required(VERSION 3.9)" in the main CMake file.

I've been using CMake 3.16.1

How should I proceed, is it ok if I guard it by modifying it to "cmake_minimum_required(VERSION 3.16)" ?

behrisch:
Thanks!

Issue #9709, opened 2021-12-01
NEMA does not work and the gui got stuck


thinker310:
Hello, I want to know how to set the configuration about NEMA file. According to the document from the documentation, I try to do it but the sumo-gui got stuckfor a while,then it auto exit. The attachment is my simulation files about it. Hope you can help me solve it, thank you very muc
[nema_test.zip](https://github.com/eclipse/sumo/files/7633750/nema_test.zip)
h!

namdre:
The value of `<param key="coordinatePhases" value="X,Y"/>` is currently mandatory even if coordinate-mode is false.

@qichaow it appears that the value of `r1coordinatePhase` and `r2coordinatePhase` are used even if coordinate-mode is false. The documentation seems to imply that the values are not needed in this case. Can you recheck the code / explain?

qichaow:
The parameter coordinatePhases defines another barrier beside the parameter barrierPhases. 

In your additional file, _<param key="barrierPhases" value="2,6;4,8"/>_ should be _<param key="barrierPhases" value="2,6"/>_ and set _<param key="coordinatePhases" value="4,8"/>_ even it's not in coordinated mode.

I notice that it's confusing. @namdre Maybe we should change the name "coordinatedPhases" to "barrier2Phases." Not sure if that's okay given that SUMO 1.11.0 is out.

namdre:
@qichaow the parameter name can stay as it is. We just need to make sure it's behavior is well documented

For every parameter we must consider the following
- is it mandatory or optional?
- if it is mandatory, can we set a reasonable default (i.e. "2,6" for coordinatePhaes)?
- if there is a default, can we detect if the default makes no sense (i.e. phase 2 or 6 are missing)?

Then:
- document for every parameter whether it has a default or whether it must always be set by the user (only mandatory parameter names are written in bold in the parameter table, the others in normal font)
- document the default value if it exists (new column in the parameter table)
- adapt the code so it generates an error (`throw ProcessError`) if a mandatory option is missing or the default cannot be used. The last point is very important since sumo has a strict "no crashing"-policy.

namdre:
thanks for #9715 (the rest of my recommendation remains valid)

qichaow:
@namdre Thanks for the recommendation. I thought the documentation was clear but maybe it was only clear to my inner self. :) I will have at least another person to read the write up before I post it in the future.
[...]

PR #10014, opened 2022-01-21
add some error handling for NEMA type controller

qichaow:
Address #9709. This PR covers most of the error scenarios that I can think of. We could add more error handling as issues occur in the future.

@mschrader15, feel free to build on it as you see fit.


namdre:
@qichaow Your error handling makes 100% of the NEMA tests fail (with different error messages). Are all our inputs faulty or are the checks too aggressive?

Issue #10013, opened 2022-01-21
NEMA should support NEMA TS1 and NEMA TS2 offsets as well as Type 170


mschrader15:
@qichaow I noticed that your implementation of the offset is Type 170 (according to [Chapter 6.3.4](https://ops.fhwa.dot.gov/publications/fhwahop08024/chapter6.htm). 

We use TS2 controllers here and I am seeing some differences in testing that I think will be hard to resolve without explicitly supporting TS2.

mschrader15:
I will take the lead on this @qichaow , just wanted to make note of it here

qichaow:
Sure, go ahead.

Qichao

qichaow:
Hi Max,

Just want to let you know that TS1/2 is a cabinet standard, not a
controller type.

Thanks,
Qichao

mschrader15:
@qichaow that it is (and sorry if my language has been unclear), but it also determines how the offset is calculated no? See [Chapter 6.3.4 here](https://ops.fhwa.dot.gov/publications/fhwahop08024/chapter6.htm) for FHWA description of the offset calculation and how it relates to TS1/2 and Type 170

mschrader15:
FYI the option I added as `cabinetType` and documented it https://sumo.dlr.de/docs/Simulation/NEMA.html#param_attributes

PR #10022, opened 2022-01-23
Implements TS2 Timing #10013

mschrader15:
Implements #10013. @qichaow this refactors some existing code to support future extension to TS1 as well. 

Adds tests and documentation. 

Small peak at the validation below. Left is SUMO's NEMA controller and right is Econolite Controller in SIL. 

image


qichaow:
Looks great! I'm fully booked today. I will take a deeper look at it later.

Issue #9989, opened 2022-01-19
NEMA should allow configurable "detector lock" for the nextPhase logic


mschrader15:
NEMA users should be able to configure whether the detector are "locked" at the begging or ending of yellow. In the Econolite I am testing against, this is actually configurable on the detector level. 

We don't currently distinguish between detector lock & the next phase algorithm. @qichaow do you think it is worthwhile to allow per-detector lock configuration in the vein of #9983 or just do a general parameter?

https://github.com/eclipse/sumo/blob/bab4b98f54ca2c2ae832a02c856e84f495844d78/src/microsim/traffic_lights/NEMAController.cpp#L926-L942

qichaow:
@mschrader15 Hi Max,

What does detector lock do?

Qichao

mschrader15:
@qichaow it basically freezes the detector states at a certain point in time before the next phase is calculated. 

On the Econolite that I am working with, there is a choice between the start of yellow or the start of red.

qichaow:
Thanks. How about we start with a general parameter for now? We can take
more parameters when we have a good GUI to hide/expand parameters.

We should probably have a priority list for adding features to the NEMA
type controller so that we can capture the most used features first.

mschrader15:
Along the same lines, this should also support latching detectors

mschrader15:
Latching detectors implemented with #10598

mschrader15:
@namdre latching is different than the lock. I don't have a use case for building in the detector lock anymore, so could close it and reopen if someone else wants detector lock options.

PR #10035, opened 2022-01-25
Implements NEMA latching detectors

mschrader15:
Implements latching detectors for NEMA (see https://github.com/eclipse/sumo/issues/9989#issuecomment-1019358508)

https://github.com/eclipse/sumo/commit/f6a0f8cd8192846f183911dbf975d408050e2183 solves an edge case where a late detector call would let 3+7 -> 3+8 -> 4+7. After the phase reaches the barrier (8), it shouldn't be able to transition "backwards".

https://github.com/eclipse/sumo/commit/288fff7e6ced59396a93b52f079a877a0514fef4 adds tests


namdre:
@mschrader, there is a merging problem that must be fixed on your end. Can you rebase your PR on the current state of main?

mschrader15:
@namdre I rebased. Let me know if you see any other issues

PR #10036, opened 2022-01-25
Update NEMA documentation

qichaow:
Added the traffic light program config for a four-leg eight-phase intersection.
Added a figure for the ramp intersection that corresponds with the existing traffic light program config.
Updated the four-leg intersection figure with different colors for better visualization.


qichaow:
@namdre , I'm closing this PR. For some reasons, github didn't recognize me as me. I'm sending you the updated files so that you can update it.

namdre:
Ok

PR #10052, opened 2022-01-26
Fix the width of the internal bike lane connection

### LINKED ISSUE #10051: Fix the width of the internal bike lane connection in an intersection
The width of an internal lane is usually the one of the successor lane.

In some cases, if the predecessor is a bike lane and the successor is a normal lane, the width of the internal bike connection will be too large.

alexpagri:
Signed-off-by: Ioan-Alexandru Stef <alexstef34@yahoo.com>

Fixes #10051

### DIFF
--- a/src/netbuild/NBEdge.cpp
+++ b/src/netbuild/NBEdge.cpp
@@ -3877,11 +3877,13 @@ NBEdge::getInternalLaneWidth(
     const NBEdge::Connection& connection,
     const NBEdge::Lane& successor,
     bool isVia) const {
-    
-    return (!isVia && node.isConstantWidthTransition() && getNumLanes() > connection.toEdge->getNumLanes()) || (
-            isBikepath(getPermissions(connection.fromLane)) && (
+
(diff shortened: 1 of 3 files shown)


alexpagri:
I will prepare an update to the code.
In order to get everything right, I suggest getting the minimum width out of the predecessor/successor lane, at least for the bike lane, taking into account that bikes are now aligned rightmost on the sublane model.

namdre:
@alexpagri I agree with using the minimum width.
Since the formula for determining width now spans several lines and is used more than once, please refactor it into a function.

namdre:
The current state doesn't compile because of the extra `NBEdge::` 

I looked at the diffs (after fixing the build locally) and found one undesirable change:

In networks where the bike lanes have default width (`UNSPECIFIED_WIDTH`), the resulting internal lanes should also have default width. 
- With your PR, the .net.xml contains the attribute `width="3.20"` for these lanes 
- before the width (being set to UNSPECIFIED_WIDTH) wasn't written

This also introduces noise when evaluation the expected test outputs.

namdre:
Thanks!

PR #10097, opened 2022-02-04
Added parameter to setPreviousSpeed

Domsall:
I use setPreviousSpeed in scenarios in which a SUMO-vehicle is controlled by another program (another SUMO instance or a third-party program like Simulink).
When a program with higher timesteps forces a SUMO-vehicle to a specific speed (with setPreviousSpeed), the acceleration is calculated as if the vehicle had only changed the speed over its last timestep (but the other simulation changed the speed over a longer period with a lower decel/accel). The results are hard accelerating/braking maneuvers and create issues with the EIDM.
Therefore, I added the parameter "timefactor" to manipulate the acceleration calculation. Old behavior is timefactor=1.

### DIFF
--- a/src/libsumo/Vehicle.cpp
+++ b/src/libsumo/Vehicle.cpp
@@ -1409,19 +1409,15 @@ Vehicle::setSpeed(const std::string& vehID, double speed) {
 }
 
 void
-Vehicle::setPreviousSpeed(const std::string& vehID, double prevspeed, int timefactor) {
+Vehicle::setPreviousSpeed(const std::string& vehID, double prevSpeed, double prevAcceleration) {
     MSBaseVehicle* vehicle = Helper::getVehicle(vehID);
     MSVehicle* veh = dynamic_cast<MSVehicle*>(vehicle);
     if (veh == nullptr) {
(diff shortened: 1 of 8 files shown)


namdre:
We have a strong policy against breaking backward compatibility. 
New parameter for existing methods should be optional 
- default argument in all clients to handle old scripts
- TraCI server code to handle old and new protocol version

Also, have you considered turning the extra parameter into 'acceleration'? rather than timefactor? The latter is more ambiguous whereas acceleration is more explicit.

Domsall:
You are absolutely right. Sorry for the dirty pull request. I hope the changes reflect what you are looking for.
I also changed the parameter to 'acceleration' as you suggested.

namdre:
Almost: TraCIServer will still complain when encountering an old client that only sends a 'speed' value. rather than 2 components.
Instead, it should handle both versions. For comparison, see how TraCIServerAPI_Vehicle::processSet handles VAR_EDGE_EFFORT

namdre:
The code looks good now. I'm waiting for the tests to run and will then merge it.

Issue #5480, opened 2019-04-16
Enhancement, Simulation: bus does not use deticated buslane, if other lanes are shorter


ZoltanBaksa:
I have a test scene, where there are 3 lanes on an edge, an the leftmost one is a dedicated buslane. In the simulation the bus uses the rightmost lane. I had to change the allowed vehicles on the rightmost and middle lane to keep the bus in the leftmost lane.
This could mean a lot of manual work for simulating public transportation in a city area

behrisch:
I don't think it is related to the length of the lane since all lanes in sumo have equal length. Probably it is rather connectivity and/or the desire to stay right. You may try to increase lcSpeedGain and decrease lcKeepRight for the busses. Maybe you can also set a higher maximum speed on the left lane to make it more attractive.

namdre:
All these are possible work-arounds but the core issue is modelling vClass based lane preference in the absence of hard restrictions. 
This would require a new lane-change motivation either as configurable part of the lane change model or as part of the strategic lane preference (updateBestLanes).
The preference could either use the existing vClass information (vehicles prefer lanes with the exact matching permissions (e.g. bike lane or bus lane) over all-purpose lanes  or it could be a configurable part of the network model. 
The lane attribute 'preferred' already exists and was supposedly introduced to deal with this (it remains unused until now).

There already exists the lane attribute 'preferred' which supposedly was invented to deal with this.

namdre:
thoughts: putting this into the bestLanes computation would imply a strategic motivation that overrrules speedGain concerns.
This interaction is not necessary (the dedicated lane should generally be faster) and may be detrimental (no overtaking if a slower bus is ahead on the bus lane). 
It may be more useful to factors this into the keepRight motivation: dedicated lanes on the left should overrule the keepRight rule.

namdre:
If the dedicated lane is on the right, keepRight motivation may be "to slow" to reflect the preference of bicycle riders to use a dedicated bicycle lane as soon as possible. 

This points to the need for a distinct lane changing motivation in the hierarchy: 
`LCA_SPEED_GAIN > LCA_DEDICATED > LCA_KEEPRIGHT`

namdre:
somewhat related to #9091 (we might define a new lanePreference item that is based on the the 'type' attribute of a lane)

PR #10147, opened 2022-02-12
Bicycles prefer bike paths

### LINKED ISSUE #10146: Bicycles' preference for bike paths
Currently, bicycles do not treat bike paths any different from other lanes in their `bestLane` calculation.

It would be more realistic to have bicycles consider bike paths better than other lanes.

dolevi-tub:
Signed-off-by: Dominik Genz d.genz@campus.tu-berlin.de

Fixes #10146


namdre:
The PR breaks three tests.
Most illustrative is test `sumo/devices/rerouting/assignment/bike-speeds`:
In the case where a bike lane ends, the bicycle must switch to the road lane (to avoid teleporting).
Generally speaking, the bike lane should only be preferred if it allows to continue the route just as well as the next best alternative lane.

Please also provide a test where your PR shows a beneficial difference in behavior (as attached zip file or separate PR).

namdre:
Upon further consideration, it might be a bad idea to solve this via the strategic lane choice code (as discussed in #5480)

behrisch:
Thanks for the PR, work will continue in #5480

PR #10219, opened 2022-02-24
Added support for known color strings in sumolib's XML color decoding

alexbrummer:
The missing support led to errors when parsing e.g. polygons with colors defined as "red", "grey", etc.

Signed-off-by: Alexander Brummer <alexander.brummer@fau.de>


namdre:
Thanks!

Issue #5735, opened 2019-06-11
Vehicles are hidden beneath their coloring bubble.


namdre:
This makes it hard for the end user to check whether 3D-Models have been loaded correctly.

namdre:
The propper solution would be to get rid of the bubble and figure out a generic way to recolor the models themselves.
A quick-fix could be to draw a flat oval beneath the vehicle rather  than enclose it in a bubble.

m-kro:
Recently I played a bit with OpenSceneGraph and managed to recolor the vehicles themselves entirely, using additional osg::Material items. If the vehicle is textured, the view scheme color is added on top (not visible in the example image). 

Example image using the [bs3dosm sumocfg](https://github.com/eclipse/sumo/blob/3665087fdf29e4e158f2a90e8ced1a2fd37e3231/tools/game/bs3d.sumocfg):
![vehicleColourScheme](https://user-images.githubusercontent.com/33294728/151866693-74197192-c45e-4070-878e-46e81de56f82.png)

Should there be a fallback solution if the base model cannot be loaded? Moving lights without any vehicle body look quite strange ;-)

namdre:
A default passenger car body as fallback would probably make the most sense.

namdre:
@m-kro Are you planning to do a pull-request for this?

m-kro:
Yes I do. Didn't have the time to implement the fallback solution yet - hopefully this week. Then I will do the pull request.

namdre:
Awesome! Thanks!

PR #10227, opened 2022-02-26
Color 3D vehicles themselves and remove bubble #5735

m-kro:
This pull request changes how vehicles are displayed in the 3D OpenSceneGraph view. It removes the bubbles previously used for color scheme purpose and instead colors the vehicles directly. 

- If the 3D model uses textures, the color is blended over. 
- If the color scheme uses transparency, the transparency applies to the whole 3D model. 
- The whole 3D model is colored, including e.g. the tyres 

Current vehicle type parameter defaults guarantee every vehicle type has a default 3D model assigned, even though a passenger car might not make much sense in some cases.


namdre:
cheers!

PR #10357, opened 2022-03-17
New Feature Road Friction

ewht:
What it is:

Basically, just an extension of the network properties by a "friction" attribute, representing the available friction between a vehicle and the road. (See publications)
-	Netloader and Netbuilder has been adapted to load "old" networks and set a default friction value of 1.0
-	Neteditor full integration
-	Full TraCI integration (get/set) via lane or edge. In case of edge, it is actually the mean friction value of all the lanes of that edge.
-	There is an "additional" implemented named "VariableFrictionCoefficient" working like the VariableSpeedSign, just for the friction value if you need it to be dynamic.
-	otherwise, change it via TraCI at runtime (that is what I usually did using matlab and python scripts for sets of simulations...)

However, it comes with two adapted CarFollowing Models, KraussFric (the default one with an adaptation) and ACCFric.
-	KraussFric mimics a Human drivers behavior of going a little slower with deteriorating Weather (= friction) conditions (see also the pubs)
-	ACCFric enhances the "safety" distance and adapts speed (if necessary) depending on the prevalent friction value. 
 Note that in both cases, they behave exactly like their parent for a value (default) of 1.0! --> That is a feature :)

Furthermore, it comes with a device (MSDevice_Seeroad) that mimics a sensor system that you can equipp to a vehicle that "measures" the value on the road.
It comes with a statistical model. For simplicity, it is only white noise. You can adapt:
-	the standard deviation "device.Seeroad.stdev"
-	and add a fixed offset (systematic measurement error) "device.Seeroad.offset"
if you want via the custom attributes.

I pulled and merged with the version from Mar 16 11 a.m. (1.12.xxx) and compiled Windows and Ubuntu 20.04 LTS both working and passing my tests.
You can find example scenarios to play and test with here: (same github as the fork) 
https://github.com/ewht/friction_example_scenarios

If you are willing to migrate, I would also start updating the docs, but if not I wouldn't bother since we are using it inhouse. 
We have a couple of additional tools (python...) that I have not integrated yet (would need to clean the code and such), but not if there is no need.
For questions feel free to contact me anytime: thomas.markus.weber@uni-due.de

References:

Dissertation: (English)
https://doi.org/10.17185/duepublico/75278
On the Potential of a Weather-Related Road Surface Condition Sensor Using an Adaptive Generic Framework in the Context of Future Vehicle Technology

Abschlussbericht: (Final Project Report (German))
https://doi.org/10.2314/KXP:1741603366
Sensorsystem zur autonomen Fahrbahnzustandserkennung (SEEROAD) : Abschlussbericht : Teilvorhabenbezeichnung: Fahrzeugmodell, Fahrzeugsimulation : Laufzeit: 1. März 2017-29.Februar 2020

SUMO User Conference 2019 (English)
https://doi.org/10.29007/cqps
Introducing Road Surface Conditions into a Microscopic Traffic Simulation


behrisch:
Thanks a lot for the monumental effort! I would strongly propose that we split this into multiple requests if possible. Maybe starting with the network related changes then the models and finally the device. Can you try to do  this? Furthermore I would suggest to make the friction attribute optional in the net input and output which would limit the number of changes drastically.

ewht:
okay, I see what the problem is with all the million test files and such. will close this one and reopen a new one once i only touch the src files. Then you'll see that it is actually not that much :) Anyhow, i see how the optional netwrite is beneficial (the load/building already is optional, as it sets NBEdge::UNSPECIFIED_FRICTION as default if none is provided). I'll change it so that it is legacy compatible.

PR #10393, opened 2022-03-22
New Feature Friction

ewht:
Update to the closed pull request #10357
1	Made sure the diffs are only the necessary ones
2	Merged with [fcb8c93](https://github.com/ewht/sumo/commit/fcb8c93b1f231716c30216e6dedc733ae62f4362) today (22.03.2022)
3	Changed netwriting to only write the attribute if it is not UNSPECIFIED (default) meaning it is now fully legacy compatible. 
	If you don’t explicitly use the friction attribute, you wont notice that it is there.
What it is:
Basically, just an extension of the network properties by a "friction" attribute, representing the available friction between a vehicle and the road. (See publications)
-	Net-loader/-builder/-importer has been adapted to use a default friction value of 1.0 if UNSPECIFIED
-	Neteditor full integration along with 
-	There is an "additional" implemented named "VariableFrictionCoefficient" working like the VariableSpeedSign, just for the friction value if you need it to be dynamic.
-	Full TraCI integration (get/set) via lane or edge. In case of edge, it is actually the mean friction value of all the lanes of that edge.

It comes with two adapted CarFollowing Models, KraussFric (the default one with an adaptation) and ACCFric.
-	KraussFric mimics a Human driver’s behavior of going a little slower with deteriorating Weather (= friction) conditions (see also the pubs)
-	ACCFric enhances the "safety" distance and adapts speed (if necessary) depending on the prevalent friction value. 
 Note that in both cases, they behave exactly like their parent for a value (default) of 1.0! --> That is a feature :)

Furthermore, it comes with a device (MSDevice_Seeroad) that mimics a sensor system that you can equipp to a vehicle that "measures" the value on the road.
It comes with a statistical model. For simplicity, it is only white noise. You can adapt:
-	the standard deviation "device.Seeroad.stdev"
-	and add a fixed offset (systematic measurement error) "device.Seeroad.offset"
if you want via the custom attributes.

You can find example scenarios to play and test with here: (the same GitHub as the fork) 
https://github.com/ewht/friction_example_scenarios

For questions feel free to contact me anytime: thomas.markus.weber@uni-due.de

References:

Dissertation: (English)
https://doi.org/10.17185/duepublico/75278
On the Potential of a Weather-Related Road Surface Condition Sensor Using an Adaptive Generic Framework in the Context of Future Vehicle Technology

Abschlussbericht: (Final Project Report (German))
https://doi.org/10.2314/KXP:1741603366
Sensorsystem zur autonomen Fahrbahnzustandserkennung (SEEROAD) : Abschlussbericht : Teilvorhabenbezeichnung: Fahrzeugmodell, Fahrzeugsimulation : Laufzeit: 1. März 2017-29.Februar 2020

SUMO User Conference 2019 (English)
https://doi.org/10.29007/cqps
Introducing Road Surface Conditions into a Microscopic Traffic Simulation


behrisch:
Thanks again for your work! 122 changed files is still a lot and I would prefer to do this step by step so that we can better reproduce if changes break things. The automated tests also did not run cleanly.

I set up a new friction branch now in the main repo. So if you do not want to do the work of splitting the PR alone (which I totally understand) we can merge the PR into the branch and continue from there. What do you think?

ewht:
Hi, sounds reasonable :) I checked on the tests and found the issue. I fixed it so that the fork is passing them now, so you might want to use the version as of today.

ewht:
> Can you do a new PR without the conflicts against the friction branch or should I try to resolve this?

did it a couple of minutes ago I guess we can close this one?

PR #10450, opened 2022-03-30
Friction

ewht:
fixing the merge conflicts due to reverting to an earlier commit last Friday.
merged all the way up to main/726ae52 from today


behrisch:
It took a while but now I merged now most of the non-GUI (and not netedit related) stuff. I followed a slightly different approach by making the availability of friction depending on the device but now it is available in every car following model, removing the need for the extra model and also the random noise is always applied (as in your seeroad device).

I hope you agree with these changes.

Furthermore I incorporated the variable friction functionality directly in the variable speed sign since it is very similar and removes a lot of duplicated code and the need for a separate traci domain.

One thing I added to the model for now is to force the old behavior if friction is set to 1. Your interpolation did not give exactly 1 for the factor if friction was 1 but now we have slight discontinuity around 1. Can we maybe to the second order fit again under the constraint of f(1) == 1?

Thanks again for the contribution and feel free to add comments here!

PR #10598, opened 2022-04-26
NEMA Controller Re-write + SUMO UC 2022 Fixes / Tweaks

mschrader15:
Big commit that I tried to split up logically:

-  https://github.com/eclipse/sumo/commit/dca4c4de55a6dcf6fd038e8a0bca83fc6ad1d9cf re-writes the NEMA controller in an object-oriented fashion. It introduces two additional classes 
    - `NEMAPhase`, which stores timing parameters, detection-relevant parameters, and a list of valid transitions
    - `PhaseTransitionLogic`, which stores the actually switching logic between two phases.
    - Both are intended to be extensible in the future
- https://github.com/eclipse/sumo/commit/5f6b093798d8a7956bccff330c00a85d088b539e updates some of the NEMA test files to match a switch in the controller default settings. I was only able to validate the controller against a TS2-type controller and so feel (along with @qichaow ) that the default `controllerType` should be TS2 going forward. Most of tests were done with a Type170 controller so I needed to update the tests to explicitly set the `controllerType`.
     - The parameter `cabinetType` was changed to `controllerType` per discussion with @qichaow 
- https://github.com/eclipse/sumo/commit/5de24d9183d86b87ba0ee5c0219edfd0b4efb503 updates test results to match the paper changes + new controller design
- https://github.com/eclipse/sumo/commit/1ede5c07849d77bda185598d55c5f66cb0cbcba8 updates the online documentation to match the new parameter naming + add the new parameters


namdre:
Thanks for update @mschrader15 @qichaow . This has come just in time for the planned release next week.

I've noticed that the error messages have changed from

```
Warning: At NEMA tlLogic '0', the phases before barrier 1 from both rings do not add up. (ring1=169.00, ring2=174.00)
Warning: At NEMA tlLogic '0', the phases before barrier 2 from both rings do not add up. (ring1=102.00, ring2=97.00)
```
to

```
Warning: At NEMA tlLogic '0', the phases before barrier 1 from both rings do not add up. (ring1=97000, ring2=102000)
```
(i.e. in tests/sumo/basic/tls/NEMA/cross_phase_switching/errors.sumo)

I've already fixed the time format but I'm still wondering why the times have been switched around for the rings and why there is only one warning now.  Can you confirm that the change is expected?

mschrader15:
So the previous ring calculation was wrong. For example in `tests/sumo/basic/tls/NEMA/cross_phase_switching/errors.sumo`:

Ring 1 is 1,2 | 3,4
Ring 2 is 5,6 | 7,8

The sum of times from `tests/sumo/basic/tls/NEMA/cross_phase_switching/input_additional2.add.xml` is:

(20 + 4 + 2), (65 + 4 + 2) | (40 + 4 + 2), (20 + 4 + 2)
(25 + 4 + 2), (65 + 4 + 2)  | (40 + 4 + 2), (20 + 4 + 2)

So it is only barrier 1 that does not sum, as ring 1 barrier 1 = 97 and ring 2 barrier 1 = 102

For barrier 2, the existing algorithm was summing the entire cycle time (169 for ring 1 instead of 72)

PR #10699, opened 2022-05-09
Display 2D decals in OSG 3D view

m-kro:
It creates a plane with the decal as a texture, then uses it like any other 3D object one could add using the decal settings. In order to avoid flickering of overlapping decals and edges, Z position can be adjusted with the layer value.

@namdre: Should rather the "altitude" attribute of a decal be used for Z position? This is currently not accessable in the GUI dialog.


namdre:
> Should rather the "altitude" attribute of a decal be used for Z position? This is currently not accessable in the GUI dialog.

Polygons also use their `layer` attribute as z-coordinate so I think it's at least consistent to do the same for decals

PR #10808, opened 2022-05-25
New unit tests for the Boundary class

vijay-jaisankar:
- Ensuring that the right parameters are updated in the `Boundary::set()` function.
- New unit tests for the `Boundary` class' `growWidth()` and `growHeight()` methods.
- Documentation conforming to the current tests' implementation.

Note:
In the `grow()`,  `growWidth()` and `growHeight()` functions, we need not check if `by >= 0` and negate `by` if not, as one can view this operation as "shrinking" the boundary.


namdre:
Thank you

PR #10824, opened 2022-05-30
Migration of the sumopy contributed tool from python 2.x to python 3

vijay-jaisankar:
- Updating the `sumopy` tool's code using the [2to3](https://docs.python.org/3/library/2to3.html) tool, so that it runs with python 3
- Detailed changes about the same are listed in the `migration_changes.txt` file (in `tools/contributed/sumopy`)


schwoz:
yes, time.

It is not a major problem to migrate, but it'll take time, because I 
need to move from wxPython 3.0 to  wxPython 4.x and this would bring 
some substantial changes to the GUI, in particular in conjunction with 
displaying PyOpenGL

I am aware of the urgency to move to Python 3 and I will do it as soon 
as work load allows. I'm also OK with removing SUMOPy from the 
repository until this has happened.

Best wishes,
Joerg

behrisch:
@schwoz But does it help if we merge this one or not?

EwoutH:
Hi! I’m looking to use sumopy with Python 3. That is currently not possible, right?

I have some experience with Python upgrading, so if desired, I can assist.

schwoz:
We have migrated to Python3! The software is now called HybridPY because it can also manage mesoscopic simulations with Matsim. There is a bit of code cleaning to be done, I will let you know when we move it to this repository.

EwoutH:
Awesome! Just finished the last simulation run of my thesis tonight (using [UXsim](https://github.com/toruseo/UXsim), among others), but will consider it for future projects!

behrisch:
#17365 is going to solve this

PR #10831, opened 2022-05-30
Python Installation Folder in `start-command-line.bat` file

vijay-jaisankar:
- Setting the python directory in the installation file


namdre:
The whole point of configuring python_dir is that people may not have it on their binary search path (and thus cannot just call 'python').
Your PR does not help with this problem.

vijay-jaisankar:
Thanks for the clarification

Issue #10773, opened 2022-05-20
walkingareas look glitched in 3D view


namdre:
@m-kro it appears they are drawn at the same z-offset as the junction shape and this creates artifacts.

namdre:
example network: tests/tools/import/OSM/webWizard/osmimport.tools

m-kro:
Currently, walkingareas are treated the same as other road segments, using their given coordinates - no special z offset is applied. 
I don't know why the other example works, but would expect it to be random / a floating point issue. So should we add a small z offset then?

namdre:
Yes. I think a z-offset that puts it above the junction would be best. also for crossings probably.

behrisch:
maybe rise walking areas 10cm similar to the real world

m-kro:
![grafik](https://user-images.githubusercontent.com/33294728/171349626-06d6ebcc-8150-4715-8157-9134608563e2.png)

- The screenshot shows walkingareas and sidewalk lanes raised to 10cm height. 
- The shading / normals are now computed by OSG itself, as the ones defined manually would brake when applying the polygon tessellator. This leads to some "rounding shadow" in some places.
- The near/far ratio of OSG has been increased: Small Z offsets are now rendered correctly. This comes with the price of very close items not being rendered anymore.
- There is currently no good solution for overlapping shapes in the same plane, like rail tracks inside a junction

namdre:
looks nice!

m-kro:
As far as walkingareas are concerned, I'd consider this issue as done. Obviously there remain other issues with overlapping shapes like tram tracks across junctions...

PR #10851, opened 2022-06-01
extrude walkingarea and sidewalk lanes in OSG ref #10773

m-kro:
- In order to remove the glitch of shapes with very short Z offset, walkingareas and sidewalk lanes (lanes for pedestrians only) have been raised by 0.1 from the ground and solidified with kerbs. 
- The OSG near/far ratio has been increased. As all object distances from the camera are normalised to [0;1], we gain some precision for close objects with a higher near/far ratio. This comes at the price of not seeing anymore items very close to the camera (approx. 2m?). Might be an issue for a driver perspective.


m-kro:
Closed because containing a few changes not relevent to the topic

Issue #10791, opened 2022-05-23
Zoom through 3D viewport dialog changes camera direction


m-kro:
When changing the zoom value in the 3D viewport dialog of sumo-gui to values > 100, the Z value of the camera location decreases, but the lookAt coordinates remain unchanged. This means the lookAt position (= the point the camera should look at) is above the camera, thus the camera points into the sky.

**SUMO-version:** current dev ver.

namdre:
first thought:  shouldn't the look-at coordinate be always on the ground?
second thought: in complex 3d networks, it may be desirable to point the camera skywards (https://en.wikipedia.org/wiki/Stack_interchange)

m-kro:
I changed the title to make the problem more clear. Explicitly looking to the sky is possible by either entering the corresponding coordinates in the dialog or manipulating the view directly. But zooming should preserve the camera direction  which currently is not the case.

m-kro:
The zoom value in 3D view is now only used as a helper within the viewport dialog, as the view is already entirely defined by the camera and the position to look at. 

- A zoom value of 100 leaves the view unchanged.
- Values greater 100 zoom in by (zoom - 100) meters, 
- Values smaller than 100 zoom out in view direction
- The zoom value is reset to 100 when the viewport dialog is shown the next time

PR #10854, opened 2022-06-01
Zoom along the view axis in OSG 3D view ref #10791

m-kro:
The 3D OSG view now recalculates the viewport coordinates (camera position and where to look at) when the zoom level is changed in the viewport dialog. The zoom level in the 3D view is defined the same way as in 2D: A zoom level of 100 is equal to a Z value of the network width and a higher zoom level means getting closer to the ground.

Caveat: If the camera is positioned in a parallel way to the ground, this concept does not work. In this case, a warning is issued.

### DIFF
--- a/src/osgview/GUIOSGView.cpp
+++ b/src/osgview/GUIOSGView.cpp
@@ -125,7 +125,7 @@ GUIOSGView::GUIOSGView(
     GUINet& net, FXGLVisual* glVis,
     FXGLCanvas* share) :
     GUISUMOAbstractView(p, app, parent, net.getVisualisationSpeedUp(), glVis, share),
-    myTracked(0), myCameraManipulator(new SUMOTerrainManipulator()), myLastUpdate(-1) {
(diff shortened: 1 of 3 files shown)


m-kro:
Currently the zoom history / the previous zoom value is not stored as it is a derived value anyway. So there is no backward/forward. I think then a different approach is better than trying to solve it using the 2D zoom level def.

The easiest thing would be to always use the view vector as a reference for the zoom level. The vector is normalized, so a bird view camera 100m above the ground would be equal to zoom level 100.

m-kro:
The alternative approach is stored [here](https://github.com/m-kro/sumo/tree/ticket10791alt):

- zoom factor 1 = unity view vector
- zoom factor _f_ = scale view vector by factor _f_
- _f_ scales the view vector within the boundaries allowed by GUI [0.0001, 100000]
- _f_ is reset to 1 when the viewport dialog is closed (should rather be invalidated by moving the OSG view by mouse, but could not get that to work)

m-kro:
Discussion about the alternative appraoch continues in the corresponding issue #10791

Issue #10848, opened 2022-05-31
POI Colour Filters for Images


vijay-jaisankar:
## Relevant Files

The following files are relevant to this conversation:

`pink.png`
![pink](https://user-images.githubusercontent.com/56185979/171292223-cb578d46-e4f8-48b6-83c6-0bf253c03287.png)

`green.png`
![green](https://user-images.githubusercontent.com/56185979/171292228-7ac8c58e-4359-4f83-bd42-f8785d165b4c.png)

## Situation Context

While initialising a POI in TraCI, the`x`, `y`, and `color` attributes are mandatory.  
If we initialise the POI with an `imgFile`, the `color` provided acts as a filter for the image. 

For example,
```py
traci.poi.add(poiID="demo100", x = 3834.7883983257707,  y = 3268.0471488538005, color=(255,0,0), layer = 202.0, imgFile="/home/vijay/Pictures/pink.png") 
```
Here, we can see the superimposition of the red in the `color` attribute and the pink of `pink.png`.
![Image Filter](https://user-images.githubusercontent.com/56185979/171289968-a476faf5-6509-42da-8061-316018349024.png)

If one wishes to use the same colour as the image provided (i.e "no filter"), they would have to pass the color as `(255,255,255)` + alpha(if desired).

For example,
```py
traci.poi.add(poiID="demo100", x = 3834.7883983257707,  y = 3268.0471488538005, color=(255,255,255), layer = 202.0, imgFile="/home/vijay/Pictures/pink.png")
```
![Raw Image](https://user-images.githubusercontent.com/56185979/171290488-ba6ea707-e222-4987-8773-de66514bbc81.png)

While this "compounding" colour aspect might be surprising at first glance, a problem occurs as the same filters are applied to any new image file for the POI (`setImageFile` arguments). Unless a user knows about this effect, they will get very confused as the colour they're expecting might not manifest in the image.

For example,
```py
traci.poi.add(poiID="demo100", x = 3834.7883983257707,  y = 3268.0471488538005, color=(255,0,0), layer = 202.0, imgFile="/home/vijay/Pictures/pink.png") 
traci.poi.setImageFile("demo100", "/home/vijay/Pictures/green.png")
```
![Changed Image](https://user-images.githubusercontent.com/56185979/171291377-ebd28411-bbcc-4dc7-a6e7-c26116abf703.png)
We can see the effect of the initial filter.

## Proposed Changes
[...]

PR #10872, opened 2022-06-03
Default POI colour with the presence of imgFile

vijay-jaisankar:
Ref: https://github.com/eclipse/sumo/issues/10848

In this PR, the default POI colour has been updated depending on the presence or absence of a non-empty `imgFile` attribute.


behrisch:
Thanks!

namdre:
Actually, there was a reason why I didn't merge this yet: The PR only works for relative imgFile paths and sets the old default color 'red' for absolute paths.

I just didn't get around to writing this up before )-:

PR #10877, opened 2022-06-06
New statistical Methods

vijay-jaisankar:
- A method to normalise the given values into a specified range
  - This can be used to generate summary plots with the values from 0 to 1, for example
- Scaling Quality Value method
  - This is an alternative to the GEH method used for error calculation which gives an output between 0 and 1
  - Documentation for the method has also been added
  - Relevant links for GEH and SQV: [Wikipedia](https://en.wikipedia.org/wiki/GEH_statistic) and [Paper](https://journals.sagepub.com/doi/10.1177/0361198119838849)


behrisch:
Thanks for this one! Please reformat it according to the style of the rest of the file (no trailing spaces, doc strings after the method header, no superfluous empty lines)

vijay-jaisankar:
Sure thing, will do.

vijay-jaisankar:
The code has been modified as mentioned, please check.

vijay-jaisankar (commented):
As I was browsing the repo, I found that this file: 
https://github.com/eclipse/sumo/blob/main/tools/sumolib/visualization/helpers.py 
has a function `linNormalise` similar to `normalise_to_range`, but over there, we pass in a dictionary.

Should we use that function here? 
\+ : avoiding re-declaration of a very similar function 
\-  : type conversions (list to dict and vice-versa) and extra imports. 

If we should change it, I will make another commit to this PR.

What do you think?

behrisch:
It is such a tiny function I don't think it makes much of a difference. Just leave it as it is. But I still see unnecessary white space changes in the diff. Can you please fix those?

vijay-jaisankar:
Okay, thanks for the clarification! 
I will remove the unnecessary whitespace now.

behrisch:
Thanks this was not the whitespace I meant but I just fixed with our automated style checker

Issue #10895, opened 2022-06-09
Crash of sumo-gui when setting up GL2 compatible shaders


RNippold:
When trying to execute a 3D simulation in sumo-gui, it crashes with to following error: 

"StatsHandler::StatsHandler() Setting up GL2 compatible shaders". 

The network is loaded successfully, I can rotate and navigate through the network. Furthermore, the command line version can execute this simulation example successfully. 

SUMO version: v1_13_0+0703-89572ed5a65
OS: linux (but crashes under windows as well)
[3d_demo.zip](https://github.com/eclipse/sumo/files/8870160/3d_demo.zip)

m-kro:
Probably related to custom 3D shapes for vehicles

RNippold:
Appeared to be a problem related to specific 3ds models...

behrisch:
nevertheless whatever strange models I feed to sumo it should not crash

PR #10902, opened 2022-06-10
Use default shape when car model cannot be loaded ref #10895

m-kro:
Since removal of the big bubbles in #10227 , there was no fallback solution when a given car model file could not be loaded. Only an error message was issued. Now a box shape is drawn for vehicles with failing car model file. This should help in cases like #10895. 

### DIFF
--- a/src/osgview/GUIOSGBuilder.cpp
+++ b/src/osgview/GUIOSGBuilder.cpp
@@ -379,7 +379,8 @@ GUIOSGBuilder::buildMovable(const MSVehicleType& type) {
     if (myCars.find(osgFile) == myCars.end()) {
         myCars[osgFile] = osgDB::readNodeFile(osgFile);
         if (myCars[osgFile] == 0) {
-            WRITE_ERROR("Could not load '" + osgFile + "'.");
+            WRITE_ERROR("Could not load '" + osgFile + "'. The model is replaced by a box shape.");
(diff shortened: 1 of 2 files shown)


behrisch:
Could we have something which still allows to identify the orientation (similar to the default triangle in 2D)?

m-kro:
Well the only predefined matching shape in this respect is a cone. I'll try it out.

src/osgview/GUIOSGBuilder.cpp
```diff
@@ -379,7 +379,13 @@ GUIOSGBuilder::buildMovable(const MSVehicleType& type) {
     if (myCars.find(osgFile) == myCars.end()) {
         myCars[osgFile] = osgDB::readNodeFile(osgFile);
         if (myCars[osgFile] == 0) {
-            WRITE_ERROR("Could not load '" + osgFile + "'.");
+            WRITE_ERROR("Could not load '" + osgFile + "'. The model is replaced by a box shape.");
```
behrisch, line 382: actually the error message is now wrong it is not a box anymore

m-kro: Was corrected in commit https://github.com/eclipse/sumo/commit/1a8d867cbe00b3991ff7e213848a1e72958452a2

Issue #10691, opened 2022-05-06
improve netedit demand-mode documentation screenshots


namdre:
A large part of the screenshots is taken up by the "message window" (many with warnings (!) in it). Cropping those images would help.
Updating the screenshots to show some newer capabilities (i.e. flow attributes) might also help.

Issue #10590, opened 2022-04-26
add missing detector tool tests


namdre:
The following should be added to https://sumo.dlr.de/docs/Tools/Detector.html  and receive tests

- [ ] filterFlows
- [ ] mapDetectors (#6613)
- [ ] plotFlows.py (should also be linked from https://sumo.dlr.de/docs/Tools/Visualization.html
- [ ] routeUsage.py
- [ ] validate.py

behrisch:
also see #1425

PR #10910, opened 2022-06-12
add: collinearity checks in NBHeightMapper

vijay-jaisankar:
In this PR, the following features have been added:
- Checking for collinearity in the method which initialises a triangle with corner vectors
- Updating the vectors in case collinearity is detected

### DIFF
--- a/src/netbuild/NBHeightMapper.cpp
+++ b/src/netbuild/NBHeightMapper.cpp
@@ -373,7 +373,18 @@ NBHeightMapper::clearData() {
 NBHeightMapper::Triangle::Triangle(const PositionVector& corners):
     myCorners(corners) {
     assert(myCorners.size() == 3);
-    // @todo assert non-colinearity
+
+    // Check and correct for collinearity (ref:https://math.stackexchange.com/a/636229/728074)
+    Position v1 = myCorners[1] - myCorners[0];
+    Position v2 = myCorners[2] - myCorners[0];
+   
+    if(v1.crossProduct(v2) == 0){
+        // Update the trace elements to make the determinant non-zero
+        myCorners[0][0] = 1.01*myCorners[0][0];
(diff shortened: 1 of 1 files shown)


namdre:
Can provide any kind of input data which can serve as a test case for your fix?

vijay-jaisankar:
Sure, in the style of `NBHeightMapperTest.cpp`, 
```cpp
       NBHeightMapper& hm = NBHeightMapper::myInstance;
        PositionVector t1;
        t1.push_back(Position(2,4,6));
        t1.push_back(Position(4,8,12));
        t1.push_back(Position(8,16,24));
        hm.addTriangle(t1);
```

This represents three points which are collinear.

After processing, the transformed points are **{2.02, 4, 6}**, **{4, 8.08, 12}**, **{8, 16, 24}**
These points are not-collinear.

namdre:
I don't think "fudging" the input data is a useful approach to handle collinearity. At least I don't see how it's useful in the context where it's being used (NBHeightmapper).  The use case of that code is applying height data from a triangular mesh to a network. The question is whether this use case is negatively impacted by  an ill-formed triangle and whether your fix has a better performance. The proposed unit test cannot be used to argue for any benefit. 
I suppose your PR was motivated by the "todo"-statement in the code but to my knowledge there was never a report of this code being problematic. Hence I think the code can just stay as it is. 
That being said, we do appreciate your eagerness to help. If you help us to understand your motivation we may be able to point you at tickets/issues where you can apply your energy more fruitfully.

vijay-jaisankar:
Thanks for the feedback.

Yes, I did make this PR when I was browsing the code and saw the to-do comment :)

As of now, I'm interested in writing more tests and documentation, so if you can direct me to some issues there, that would be great.

namdre:
Here are some suggestions: #10590, #10691

PR #10951, opened 2022-06-16
Update Docstrings (TraCI)

vijay-jaisankar:

- TraCI's functions' docstrings updation (where the function signatures/names were inconsistent) with the method body


namdre:
awesome!

Issue #10942, opened 2022-06-15
netdiff.py: support saving and loading configurations


namdre:
netdiff.py uses positional arguments.  What would be a good way to include these in a config file?

behrisch:
The argument parser supports names for positional args already. So instead of parsing only the known args we could give names to the positionals and use them in the config

namdre:
Effectively this would provide an alternative to pass positional args as named args by using those names, right? That would be a good thing in general for all our position args.

behrisch:
yes

vijay-jaisankar:
![Positional Arguments](https://user-images.githubusercontent.com/56185979/174124432-1e3c2512-f7a7-4b61-91f4-50f646855650.png)

As per the above analysis, these files use positional arguments, we can update these to make them use named arguments. 
If this seems like a good idea, please assign me to the same.

behrisch:
not every script which refers to sys.argv has positional arguments and this ticket refers more to the ArgumentParser being used to parse the only some of the arguments but if you feel like it please go ahead

namdre:
before there is another misunderstanding, **we are striving for backwards compatibility here**. Positional arguments should still be usable but there should also be a way to use named arguments instead.

behrisch:
Actually it is not about new options at all it is just about the parsing code. The argument parser can give the positional argument an _internal_ name which should be used in the config as well.

vijay-jaisankar:
I was thinking of doing something similar to the scheme in `tools/game/racing.py` (with special consideration to any default arguments used in the functions.) Would that be okay?

behrisch:
No, actually not. argparse supports both "real" options starting with "-" or "--" and positional arguments which get an internal name see https://docs.python.org/3/library/argparse.html#the-add-argument-method. My proposal was to replace the positional arguments which are currently extracted from the remaining args by the named args of argparse

vijay-jaisankar:
Thanks for the info, so from what I understand, if we have something like
```py
def main(in, out, opt=10):
  # Function body
```
The parser would be like
```py
parser.add_argument('in', ...)
parser.add_argument('out', ...)
parser.add_argument('-opt', ...)
```

Is this correct?

behrisch:
yes although Iwould not state it in terms of the main function but rather how the script can be called from the command line. So it needs to be `script.py infile outfile` with an optional `-opt` switch

vijay-jaisankar:
Okay, thanks for the clarification!

[...]

PR #10975, opened 2022-06-18
add: argparse for positional and optional arguments ref: #10942

vijay-jaisankar:
ref: #10942 

In this PR, the following changes have been added:
- Command Line Arguments Parsing with `argparse`


vijay-jaisankar:
Sure thing, will change the lines with fstrings to normal `%` prints.
Sorry, I got tied up with a bit of work so I couldn't work on this for the last 2 days, will work on it ASAP.

Checklist, as of now:
- [ ] Use the parser in the repository (`sumolib.options.ArgumentParser`)
- [ ] fstring change

Also, as a sidenote, there are already 4 commits in this PR that fail the tests and there will be at least 2 more with these changes, so can I make a new PR for this when the changes are done (with only one commit)?

behrisch:
Yes please do a new PR and maybe even a new issue because it really has a scope beyond #10942. While you are it, you can possibly also remove the `if len(args) < ...` parts then because the checking is now done by the argument parser. Thanks!

PR #11010, opened 2022-06-22
New LC Parameter Overtake DeltaSpeed threshold 

Domsall:
Referred to in issue #10954:

> 3. I actually changed a parameter of the lane change model (in the Code) back then to overcome these issues with late lane changing. I will try to make it available soon.

Vehicles with high values of this parameter (3-5) will drive more defensively and brake more often to change a lane than overtake a leader.

Does it make more sense to make this parameter a multiplier?
E.g. `dv-threshold = lcOvertakeEagerness*vLaneMax` instead of `dv-threshold = lcOvertakeDeltaSpeed`


behrisch:
Does this need netedit support? Otherwise it looks just fine to me

Domsall:
I will add netedit-support (also for KeepRightAcceptanceTime).

What about the multiplier-question? I think it may make more sense.
Example:
- Right now with lcOvertakeDeltaSpeed=3.0: We would always compare with 3.0. If the speedlimit is 13.89, then the difference between vehicles may not be that high than with a speedlimit of 33.33.
- With a multiplier of 0.1: On a street with a speedlimit of 13.89, we would get 1.389 m/s as lcOvertakeDeltaSpeed. On a street with a speedlimit of 33.33, we would get 3.33 m/s as lcOvertakeDeltaSpeed.

behrisch:
I am not sure here. My impression is that people look more at the speed difference or more precisely at the time they will need for overtaking so the length of the vehicle they overtake plays a more important role than the speed limit on the street. Maybe there is a study and we can find a reference here?

Domsall:
The length of the vehicle / distance to overtake is already taken into account.
My assumption was following:
- When I add a constant parameter for all streets and speedlimits, the vehicle may try to overtake on highways, but seldom in the city.
- When I add a speedlimit dependent parameter, overtaking in the city would be similar dependent on the time to overtake than on highways.

A speedlimit dependent parameter would correspond to the speedfactor.

Domsall:
I added netedit support and changed the speed difference value into a speed difference factor for further review.

Domsall:
Will this make it into the next release?! Is something missing/failing?

behrisch:
No, unfortunately we did not find the time to properly review and test it. Sorry about that.

namdre:
Thanks for opening up another hard-coded constant to user configuration!

Issue #11018, opened 2022-06-23
Improving argparse support


vijay-jaisankar:
Relevant issues: https://github.com/eclipse/sumo/pull/10942
Base code (changes listed below): https://github.com/eclipse/sumo/pull/10975

Relevant files: ![files](https://user-images.githubusercontent.com/56185979/174124432-1e3c2512-f7a7-4b61-91f4-50f646855650.png)

Task list:

- [x] Use the parser in the repository (`sumolib.options.ArgumentParser`)
- [x] `fstring` : change to `%`
- [x] Remove `if len(args) < ...` clauses as they are not required.

PR #11024, opened 2022-06-23
Positional and Optional Arguments (Ref: #11018)

vijay-jaisankar:
Relevant issues and PRs: #11018 #10942

- Usage of `OptionParser` for positional and optional arguments 
- Special care has been taken not to use `f`strings (to maintain compatibility with python 2)

### DIFF
--- a/tools/output/vehrouteDiff.py
+++ b/tools/output/vehrouteDiff.py
@@ -29,13 +29,12 @@ from sumolib.options import ArgumentParser
 
 
 def parse_args():
-    USAGE = "Usage: " + sys.argv[0] + " <orig> <new> <out> [--earliest]"
     optParser = ArgumentParser()
     optParser.add_argument("orig", help="Original XML File")
     optParser.add_argument("new", help="New XML File")
     optParser.add_argument("out", help = "Output File")
     optParser.add_option("--earliest", action="store_true",
-                         default=None, help="Earliest Output Parameter (Default: None)")
+                         default=False, help="Earliest Output Parameter (Default: False)")
     options = optParser.parse_args()
     return options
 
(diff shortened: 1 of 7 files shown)


behrisch:
Please do not move the call to argument parsing into the main function. This makes it harder to reuse the scripts functionality as a module. But moving the parsing into a separate function is fine (although sometimes maybe a bit overkill)

vijay-jaisankar:
Sure thing, thanks for the feedback.

Please check if this new commit is okay, the `parse_args` method has been moved out of the `main` functions.

behrisch:
Yes, thank you. Another point is the USAGE string. You do not use it so there is no point in defining it.

tools/output/vehrouteDiff.py
```diff
@@ -25,14 +25,26 @@
...
 
 
+def parse_args():
+    USAGE = "Usage: " + sys.argv[0] + " <orig> <new> <out> [--earliest]"
+    optParser = ArgumentParser()
+    optParser.add_argument("orig", help="Original XML File")
+    optParser.add_argument("new", help="New XML File")
+    optParser.add_argument("out", help = "Output File")
+    optParser.add_option("--earliest", action="store_true",
+                         default=None, help="Earliest Output Parameter (Default: None)")
```
behrisch, line 38: the default should be False here

vijay-jaisankar:
Sure thing, in the most recent commit, I have removed those lines and changed the default value you mentioned in the PR feedback

tools/output/vehrouteDiff.py
vijay-jaisankar: Updated here: Commit 68a8cd4be1837f0a1d27d853da6ccabed4628210

behrisch:
Thanks a lot! There are some minor style issues but I will fix them myself.

PR #11069, opened 2022-06-27
Boundary Checks for HSV Colour Values, Addition to `<<` operator

vijay-jaisankar:
In this PR, the following functionalities have been implemented:
- Adding the `invisible` "built-in" colour to the custom `std::ostream` method of `RGBColor`
- Range checks for HSV colour values passed into the `fromHSV()` method, akin to checks done in the `interpolate` method


behrisch:
thanks!

Issue #10973, opened 2022-06-17
randomTrips.py: patch comment header of output files processed by duarouter


namdre:
when running randomTrips with option **--validate** the trips are processed by duarouter and the fact that they were created by randomTrips.py is no longer retrievable from their header. (The same applies when generating route output with option **-r**)

behrisch:
@bcoueraud87 for you

namdre:
@bcoueraud87 only the part **"The same applies when generating route output with option -r"** is still missing.

namdre:
@behrisch Do you see a feasible way of testing this with texttest? I mean, can we overwrite the **run_dependent_text** setting for a single test? We still need to filter out the timestamp but could keep the rest of the  header comment since it's all relative paths.

behrisch:
No, I don't think so but just try it out

PR #11084, opened 2022-06-28
[tools/randomTrips] Patches the trips file when the --validate option is used with randomTrips (#10973).

bcoueraud87:
The ´--validate´ option calls duarouter in the background to validate the trips. This resulted in the randomTrips configuration being overwritten in the trips file by the duarouter one. Now both configurations are inserted.

### DIFF
--- a/tools/randomTrips.py
+++ b/tools/randomTrips.py
@@ -27,6 +27,7 @@ import bisect
 import subprocess
 from collections import defaultdict
 import math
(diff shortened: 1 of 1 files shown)


behrisch (commented):
I would in fact prefer if we can do this without reading everything into memory. So just read the first line, write it out and the write the header directly into the file and then the rest

Issue #11055, opened 2022-06-26
fcd-output filtering using loaded shapes


mschrader15:
I would like to filter fcd-output for only vehicles that are "on top of" a E3 detector. My use case is to compare simulation trajectories to those recorded via a camera & radar in the same network. This is very similar to `--fcd-output.filter-edges`, which I am currently using. Being able to record just trajectories on top of E3 detectors would allow me to skip post-processing filtering of edge position and lane #. I would hope that it would be applicable to the broader community as well.

behrisch:
The problem here is that "on top of E3" is not always well defined because there may be vehicles which start inside the area but will not get detected because they did not cross an entry. But if you are fine with that I think it boils down to switching on the fcd device whenever you cross an entry and switching it off when you cross an exit.

namdre:
Not quite as simple: E3 detector-areas may overlap so you need to keep a stack of passed entries and only disable the device after clearing the stack.

namdre:
As an alternative: might it be more convenient to define a filter in terms of loaded shapes? This would also provide an easy way for users to filter lanes and positions?

namdre:
utils/geom/PositionVector provides the methods `around` and `overlapsWith` that might be suitable for this purpose. Checking the front and back vehicle positions for being within the shape (with `around`) should be the most efficient approach.

A new option **--fcd-output.filter-shapes**  that sets a list of shape ids (already loaded from additional files) might be appropriate.

mschrader15:
@namdre before I proceed with test cases (outside of my own local tests) can you see if https://github.com/UnivOfAlabama-BittleResearchGroup/sumo/commit/f608426e7823247dcba8f6d734b877baaf897c3b is inline with what you were thinking?

Could be made more efficient by testing the polygon the vehicle was most recently in first instead of a simple loop, but I didn't want to add new variables to the vehicle class to track.

namdre:
Basically, looks good.

- I would store the shapes by value rather than by pointer (since they might be deleted via TraCI)
- I would move the `dynamic_cast` int `shapeFilter` because the calling code doesn't make it obvious that a casting failure will be handled.
- If efficiency becomes a problem (with many loaded shapes) the proper technical solution is an RTree. However, I wouldn't worry about this as long as there is no practical need for filtering with a lot of shapes.
- Please also add a test case (-:

PR #11088, opened 2022-06-29
Implements #11055

mschrader15:


mschrader15:
Building the tests, I found that when a route file is loaded as an additional file, the FCD devices are created before the polygons are loaded. I switched to "lazily" creating the filters on the FCD exports first attempt to write to the file.

PR #11105, opened 2022-07-01
Update EIDM

Domsall:
Updates following open issues and netedit support

Refactored SlowtoStart-Function


namdre:
various test cases involving EIDM are failing now. Please confirm that their new behavior is expected.

Domsall:
thanks for your review. I just checked all tests and yes, the new behavior is expected.

Shall I run the tests on my Windows machine and push the results?!
What we also could do: Change the sigma-parameters to 0. Future tests should then run independent of the OS.

namdre:
If the feature under test doesn't require sigma, then I recommend setting it to 0.  We also do this for many of our tests.
Go ahead and push the expected test results afterwards.

Domsall:
A review would be great, as I am updating tests for the first time.

namdre:
looks fine to me

Issue #11110, opened 2022-07-03
Proposal for new tests


vijay-jaisankar:
I wish to add new checks for edge cases in the `unittest/src/utils/common` directory:
Here is a short description of the same: 

**FileHelpersTest.cpp**
- [x] Input of `stdout` to `test_method_fixRelative`

**RGBColorTest.cpp**
- [x] Input of `red` to `test_parseColor` 

**StringUtilsTest.cpp**
- [x] Input of `normal_string` to `test_method_convertUmlaute`
- [x] Input of `-0` to `test_method_toTimeString`

If you feel like this is a good idea, please assign me to the same.

Thanks,
Vijay

vijay-jaisankar:
The tests have been added in #11126 

Changes from description: `red` -> `RED` (as `parseColor` converts the inputs to lowercase anyways)

PR #11126, opened 2022-07-05
New test cases (#11110)

vijay-jaisankar:
- New test cases (ref: #11110)


vijay-jaisankar:
Workflow failure 

It appears that this is the issue that is causing the workflow run to fail.

behrisch:
Thanks!

Issue #11166, opened 2022-07-12
[tools/randomTrips] Create a function that checks that a given vehicle class actually exists.


bcoueraud87:
At the moment the vehicle classes in Python are manipulated with strings, which is prone to errors. Using the set `SUMO_VEHICLE_CLASSES ` defined in `tools/sumolib/net/lane.py` file, the goal of this ticket is to create a function that will check a given string corresponds to an officially registered vehicle class.

PR #11189, opened 2022-07-14
[tools/randomTrips] Added validation for vehicle classes (#11166).

bcoueraud87:


namdre:
> Some small remarks here:
> 
>     * I like the idea of throwing an exception instead of calling sys.exit. We should do this in the other places as well (but probably not in this PR)

We discussed this a bit more and came to the following agreement: 
- utility functions should generally not call `sys`.exit but rather raise an exception
- main methods should generally catch these exceptions and provide the end user with an error message but no stack trace

Issue #11199, opened 2022-07-14
IDM: collision due to imprecise braking when configured with low 'accel'


namdre:
related to #11183

namdre:
the lower accel is set, the larger the minGap-violation gets.

namdre:
The mingap-violation-factor also depends on the actual size of the minGap.
- At the default accel value of 2.6 and minGap 2.5, the actual stopping gap is 2.46 (slight violation)
- At the default accel value of 2.6 and minGap 2.0, the actual stopping gap is 2.00 (no violation)
- accel=0.2, minGap=2  -> stopGap=0.83 (triggers collision warning at factor 0.5)
- accel=0.2, minGap=1  -> stopGap=0.51 (violation but no collision warning at factor 0.5)

namdre:
From the above analysis, I conclude that the IDM model cannot be relied upon to keep even 50% of it's minGap under all circumstances. After testing somewhat extreme but not wildly implausible parameters, I'm lowering the threshold for a collision warning from 0.5 to 0.1 of minGap.

Issue #11184, opened 2022-07-13
EIDM: collision after ignoring junction blocker


namdre:
found in the same scenario that also exhibited #11165. 
The problem is related to the `backOnRoute` code at
https://github.com/eclipse/sumo/blob/26a4344f6d5873fa3e315bea65a94ccd0b515d08/src/microsim/MSVehicle.cpp#L2905-L2935

Possibly, the fact that the collision has only been observed with EIDM is a side effect of imprecise braking. The code should allow triggering collisions with any model.

namdre:
seems to be related to **--ignore-junction-blocker**

Issue #11183, opened 2022-07-13
EIDM: imprecise stopping causes emergency braking at junction


namdre:
found while building tests for https://github.com/eclipse/sumo/issues/11165 but probably not related.

The vehicle that does emergency braking is running with Krauss (after an EIDM vehicle is apparently violating it's right-of-way). ~~Possibly, some code that estimates the time to clear the intersection returns invalid results.~~

namdre:
@domsall The junction model is telling EIDM to brake hard for a closed link (oncoming foe traffic).
However, the model insists on reduced braking (~ 1/3 of the needed deceleration) and thus passes the stop line (of the internal junction).  Here is the relevant log that shows how the gap approaches 0 but the vehicle just keeps on driving.

```
stopSpeed 99.10s, veh=0 speed=2.68 gap=2.48 vars->v0_int=8.00
99.10 EIDM::finalizeSpeed s, veh=0 oldV=2.68 vPos=2.64 _vPos=2.64 vStop=2.64 vMinEmergency=1.78 vMin=2.23 vMax=2.64
stopSpeed 99.20s, veh=0 speed=2.64 gap=2.21 vars->v0_int=8.00
99.20 EIDM::finalizeSpeed s, veh=0 oldV=2.64 vPos=2.58 _vPos=2.58 vStop=2.58 vMinEmergency=1.74 vMin=2.19 vMax=2.58
stopSpeed 99.30s, veh=0 speed=2.58 gap=1.95 vars->v0_int=8.00
99.30 EIDM::finalizeSpeed s, veh=0 oldV=2.58 vPos=2.53 _vPos=2.53 vStop=2.53 vMinEmergency=1.68 vMin=2.13 vMax=2.53
stopSpeed 99.40s, veh=0 speed=2.53 gap=1.70 vars->v0_int=8.00
99.40 EIDM::finalizeSpeed s, veh=0 oldV=2.53 vPos=2.46 _vPos=2.46 vStop=2.46 vMinEmergency=1.63 vMin=2.08 vMax=2.46
stopSpeed 99.50s, veh=0 speed=2.46 gap=1.45 vars->v0_int=8.00
99.50 EIDM::finalizeSpeed s, veh=0 oldV=2.46 vPos=2.40 _vPos=2.40 vStop=2.40 vMinEmergency=1.56 vMin=2.01 vMax=2.40
stopSpeed 99.60s, veh=0 speed=2.40 gap=1.22 vars->v0_int=8.00
[...]

Issue #11182, opened 2022-07-13
EIDM emergency stop at traffic light


namdre:
found while building tests for #11165 but probably not related.

Might be caused by imprecise driving near lane end (same issue as #11183)

PR #11242, opened 2022-07-21
Fix multiple EIDM stop issues

Domsall:
Fixes EIDM issues with minGap stopping in followSpeed (see #11199) and stopSpeed (#11182 and #11183).

Updated all relevant tests.

One "EIDM"-test is failing and seems to be a different problem: #11184.


Domsall:
I found out why the last test is failing:
The calculation is based on followSpeed, so the EIDM / IDM does not come to a perfect stop behind the leading vehicle. When the ego-vehicle overshoots this position, it just ignores the leading vehicle in front.

namdre:
Here is what I found when looking at the test scenario ahead of the first collision:
Up to step 199.2, vehicle '0' decelerates (though a bit sloppy so it undercuts the minGap.
In step 199.3, the method `stopSpeed` is called in line https://github.com/eclipse/sumo/blob/23be5b3f255ef7812f8273bc092a966f27ea351f/src/microsim/MSVehicle.cpp#L2935

With input arguments of speed= 0.29 and gap=-15.42 the returned value is 0.41 and thus the vehicle starts to accelerate.
I would expect the vehicle to continue decelerating whenever a negative gap is given.

Can you check this?

Domsall:
I must have assumed to always get "positive" gaps ("negative" gaps would mean a collision for me).

The test should run successful now.

Still, is there another way to solve this issue with a blocking leader on a junction?! (Actually belongs to #11184)
When I use Krauss or IDM instead, I also get an emergency braking vehicle.

namdre:
re negative gap:  these aren't physical gaps but rather 'physicalGap - minGap', hence any mingap violation due to imprecise driving will result in a negative value.

I'll look into the emergency braking. Basically the use of option **--ignore-junction-blocker** creates *exceptional* situations where foe vehicles are somewhat ignored.  I agree that there is still a bug in sumo causing a discontinuity in behavior.  At least now all models can handle this without colliding (-:

Issue #11100, opened 2022-07-01
show vClass-specific speeds in lane parameter dialog


namdre:

vijay-jaisankar:
Can I please work on this? 

To my understanding, we should add new `mkItem` calls in `GUIParameterTableWindow*
GUILane::getParameterWindow()` of the `src/guisim/GUILane.cpp`. Is this correct?

namdre:
Yes and yes. The new items should only be shown if the lane has vClass-specific speeds. The attribute name should be something like:
```
maxspeed [m/s]         |  27.78
    maxspeed (truck)   |  16.67
```

with a slight indent to make it from a group with the default `maxspeed [m/s]` item

vijay-jaisankar:
Hello, while working on this, I faced an issue.
This is the code so far:
```cpp
 // vClass specific speed
    GUINet* net = MSNet::getInstance();
    if(net->getRestrictions(myEdge->getEdgeType()) != nullptr){
        for (const auto& item : net->getRestrictions(myEdge->getEdgeType())) {
            const SUMOVehicleClass svc = item.first;
            const double speed = item.second;
            // Todo: Convert svc into a string
    }
```

Can you please let me know if 
- I'm on the right track
- If yes, how do you convert the `svc` object into a string for the `mkItem` call? I tried `c_str()` but it didn't work.
- If not, which other function call I should focus on instead.

Thanks,
Vijay

behrisch:
You can simply use `toString(svc)` to convert it into a string

vijay-jaisankar:
Thanks a lot, https://github.com/eclipse/sumo/pull/11243 contains the changes mentioned, please check.

namdre:
Thanks!

PR #11243, opened 2022-07-22
vClass-specific speeds in the lane parameter dialog of SUMO-GUI (ref #11100)

vijay-jaisankar:
ref: https://github.com/eclipse/sumo/issues/11100

- Showing vClass specific speeds in the dialog box
- Changing `maxspeed` to `allowed speed`


behrisch:
Thank you! Please use the new c++ loop style `for (const auto& elem : map)` and do not make an unnecessary copy of the map (store it as a reference or do not store it at all). Also please change maxspeed to allowed speed for the existing entry as well

vijay-jaisankar:
Sure thing, will update the PR soon.

vijay-jaisankar:
These commits should fix the issue, please check.

vijay-jaisankar:
Thanks, I'll use this code formatting for the next PRs :)

vijay-jaisankar:
I've since validated my ECA status.

Issue #11258, opened 2022-07-27
unknown options are no longer reported


behrisch:
`~/sumo/tools/randomTrips.py --blubber blub -n ~/sumo/tests/sumo/net.net.xml` just runs but it should complain

behrisch:
The breaking commit was d7e05e1d722333b670c1509e3c9b7ec2e34a3784

behrisch:
This is not completed yet. The current option parsing is still too generous because it still accepts an option like `--foo-bar baz` but at least not `--foo baz` any longer. I would propose that the call to parse_args (or the constructor of the ArgumentParser) gets another parameter which defines the program prefixes to accept. Alternatively (and less preferred) we could hardcode the common SUMO programs (like sumo, duarouter etc.) Please discuss!

namdre:
I'd prefer an extra option for ArgumentParser.

bcoueraud87:
@namdre I would like to solve this issue.

namdre:
fixed in #11368

namdre:
also creates problems with duaIterate.py (but a different error)

PR #11368, opened 2022-08-16
[tools/randomTrips] randomTrips now reports unrecognized arguments (#11258).

bcoueraud87:
randomTrips now reports unrecognized arguments that would have been forwarded otherwise.


behrisch:
@bcoueraud87 I do not like this solution. IMHO parse_known_args is supposed to parse only the known args and should not throw an error if it encounters something unknown.

PR #11445, opened 2022-08-26
Add edge permission check to getRandomStreet

cwang717:
- add a private member `passengerStreets`(streets that allow `passenger`-type vehicles) to `AGCity`
- make `getRandomStreet()` randomly sample from `passengerStreets` instead of `streets`

`getRandomStreet()` is used in generating origins/destinations for free time trips and random background trips. While these trips are for `passenger` vehicle type by default, the returned streets can be edges that do not allow `passenger` vehicles. This PR is to fix this problem.

### DIFF
--- a/src/activitygen/city/AGCity.cpp
+++ b/src/activitygen/city/AGCity.cpp
@@ -32,6 +32,7 @@
 #include <map>
 #include <iomanip>
 #include <utils/common/RandHelper.h>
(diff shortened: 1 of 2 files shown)


cwang717:
I have registered on eclipse and signed ECA with the same email account as my GitHub account. Any idea why my ECA sign-off status does not get updated?

namdre:
Thanks!

Issue #11364, opened 2022-08-16
EIDM: vehicle does not reach stop, causing infinite sim


namdre:
the vehicle stops  20cm before reaching the assigned stop position and thus the stop never qualifies as "reached".
Consequently, the simulation does not terminate.

same tests as for #11361  (seeds 14 and 36, and no **--end** option)

namdre:
@Domsall function stopSpeed needs driving with increased precision.

Domsall:
How can I spot the erroneous stops?!

I used the following command, but was not successful:
`sumo-guiD.exe --xml-validation never --write-license --default.speeddev 0 -n net.net.xml -r input_routes.rou.xml --no-step-log --step-length 0.1 --seed 36 -e 10000` and
`sumo-guiD.exe --xml-validation never --write-license --default.speeddev 0 -n net.net.xml -r input_routes.rou.xml --no-step-log --step-length 0.1 --seed 14 -e 10000`

Issue #11361, opened 2022-08-15
creeping collision with EIDM model


namdre:
test coming soon

namdre:
@domsall  I've found collisions in 9 of 64 test cases when running with (without option --end)
`tools/runSeeds.py -k test.sumocfg -a sumo --threads 16 --seeds 0:64`

Domsall:
I found the issue and will upload it soon.
While working on this issue, I found two new bugs, one I fixed myself, the other one has to do with startupDelay:

When a vehicle is stopped, the stop-time does not count as waiting time. So during a stop, `myTimeSinceStartup` does not get reset. Could you please add this functionality? Thanks in advance!

namdre:
You mean, a vehicle that is coming back from a stop should never experience startupDelay? Why not?

Domsall:
no, it should :-)
this is just not happening right now

namdre:
understood: #11374

PR #11458, opened 2022-08-29
Issue11361

Domsall:
Solves #11364 and #11361.
Updated tests and also found another bug resulting in faulty internal speedlimits.

Please check your tests if all issues have been solved.


Domsall:
something went wrong with my first commit... What is the best way to fix this?!

PR #11460, opened 2022-08-29
Issue11361

Domsall:
Solves #11364 and #11361.
Updated tests and also found another bug resulting in faulty internal speedlimits.

Please check your tests if all issues have been solved.

EDIT: changed faulty author


namdre:
works better than before, but please see #11461

Issue #11265, opened 2022-07-30
EIDM strange acceleration behaviour


Asocius:
With timestep 0.5, everything is fine.
With timestep 0.2 or below, the bus vehicle makes strange acceleration behaviour, see image and files below.
Any idea, why this could happen? The action-step-length doesn't work with this model, so I don't know how to get a resolution smaller than 0.5s without that strange behaviour

SUMO-version: 1.13

![acceleration_behaviour](https://user-images.githubusercontent.com/66356314/181906261-d2b845af-9c94-4e1e-8ff1-cefdb2d533dd.png)
[frankfurt.zip](https://github.com/eclipse/sumo/files/9225954/frankfurt.zip)

m-kro:
@Domsall Maybe of interest for you, too.

Domsall:
Yes, I already saw it, but won't be in office until september.

What I can already tell you: The EIDM has received a lot of updates lately. Some are still open as pull request and some issues were already fixed with 1.14.1 (one issue was startupDelay, driving at low speeds with low timesteps).

When I am back in office, I will check out the updated SUMO version. I will then look into your scenario and fix the issue, if it still exists.

Domsall:
The issue is somewhat similar to #11461:
In
https://github.com/eclipse/sumo/blob/b1b3affb711ba6944307d7d4fb5eaf9adf5bed6f/src/microsim/MSVehicle.cpp#L2363
and
https://github.com/eclipse/sumo/blob/b1b3affb711ba6944307d7d4fb5eaf9adf5bed6f/src/microsim/MSVehicle.cpp#L2369
the `freeSpeed/stopSpeed` calculations update a future speed value. The EIDM assumes that the call to `followSpeed/freeSpeed` with the lowest calculated acc-value will result in the next speed update. This does not hold here and the model "gets confused". The saved values between two actionsteps do not belong to the correct `followSpeed/freeSpeed`-calculation.

@namdre:
- Can the values for `adaptLeaveSpeed` be picked up elsewhere?
- `endPos` is the position on the future lane without `seen`. Can `newStopDist` be used here?

namdre:
> endPos is the position on the future lane without seen. Can newStopDist be used here?

The computed speed anticipates the point in the future where the lane with the stop is being entered. Thus, 'endPos' is the correct distance value here. 

> Can the values for adaptLeaveSpeed be picked up elsewhere?

[...]

Issue #11106, opened 2022-07-01
AdaptToLeader calculates leaveSpeed depending on one time step


Domsall:
In the following example, `vehicle_0` is on a minor link and would like to drive onto the junction `2927559855`.
The gap between `vehicle_2` and `vehicle_3` should be large enough for `vehicle_0` to do this, but `vehicle_0` actually waits for `vehicle_3` to pass.

What I found out when debugging: 

- When the vehicle in front (`vehicle_1`) does not exist, `vehicle_0` has not problem crossing the junction.
- The `leaveSpeed` calculation without a leader results in a well estimated value and is done here: https://github.com/eclipse/sumo/blob/31b2a739c7a6751f50a8718222ffb22869243f05/src/microsim/MSVehicle.cpp#L2710
- When we have a leader on the next lane, the while-loop runs a second time and updates `leaveSpeed` in `adaptToLeader` with a value from `followSpeed`:
https://github.com/eclipse/sumo/blob/31b2a739c7a6751f50a8718222ffb22869243f05/src/microsim/MSVehicle.cpp#L2929

This issue probably only gets exposed in subsecond simulations.

A test case showing this issue: [AdaptToLeader_LeaveSpeed.zip](https://github.com/eclipse/sumo/files/9030536/AdaptToLeader_LeaveSpeed.zip)

**SUMO-version:** 1.13.0 and 1.10.0

**operating system:** Windows 10

Domsall:
I actually uploaded a test as zip in the previous comment :)

namdre:
The problem is described in https://github.com/eclipse/sumo/issues/11265#issuecomment-1231485209

Domsall:
The issue was unfortunately not resolved for my use case.

The calculated `estimatedLeaveSpeed` (in the above example Line 2710) is approx. 6 m/s. This is used when `vehicle_1` does not exist. Result: `vehicle_0` drives over the junction.

The calculated `futureVSafe` (or the old `vsafeLeader` in Line 2929) uses the current speed or the speed after one time step to update LeaveSpeed, which is approx. 0.25/0.5 m/s. Result: `vehicle_0` does not drive over the junction (because of high `leaveTime` in `MSLink::blockedByFoe`).

namdre:
Even test test I added for this issue still  shows the problem  (though many other tests show improved traffic).

PR #11560, opened 2022-09-08
New Parameter CalcReason for stop/free/followSpeed

### LINKED ISSUE #11461: EIDM: emergency braking when approaching stop
same test as as #11361,
happens on average once per simulation

quick replication with seed 38

most likely related to the new stop approach code from #11460

Domsall:
Parameter indicates what the reason for the call is and what the return speed will be used for.

Fixes #11461 and #11265

`CalcReason` was the best name I could come up with... feel free to change it :)

adaptToJunctionLeader-Code has changed a bit, but this does not change any procedure.
The new function looks like it has duplicate calls, but they differ slightly. The Code changes even result in a lower amount of stopSpeed calls.


Domsall:
Two tests were failing, updated values:

- IDM test with driverstate-device: The changed followSpeed-calls in `insertionFollowSpeed` already result in a saved "myLastPerceivedSpeedDifference"-value that is used in future calculations. If that is not wished, `pred` needs to be `nullptr`.
- FCD Output with CACC-model: In the first call (to check for insertion), caccVehicleMode is not set yet.

namdre:
1. why does `maximumSafeFollowSpeed` need a `usage` parameter? It's not virtual and should not be of concern to an cfmodel implementation.

2. can the changes in `adaptToJunctionLeader` go into a separate pull request? Everything else I understand (and it should basically have no impact on anything except for EIDM). - Having no tests except EIDM tests change would be a good sanity check here for me.

 The changes `adaptToJunctionLeader` I don't undestand (yet) and I'd like to have a detailed look at their impacts in isolation.

Domsall:
1. You added `gComputeLC=true` in #11106. `maximumSafeFollowSpeed` uses `gComputeLC` and may get called by some models in `followSpeed`. For the same output, I needed to add the `usage` parameter to `maximumSafeFollowSpeed` to make the Code work the same way as with `gComputeLC` at this point.

2. The changes to `adaptToJunctionLeader` do not change anything for other models (no tests changed and the "logics" of the Code (should) have stayed the same).

I do not know any way to "easily" undo this. Do you have an idea? The changes are in [dbc23c5](https://github.com/eclipse/sumo/pull/11560/commits/dbc23c5af05246a980ec59d4efc9721d57297e32)

namdre:
1. The gComputeLC in MSVEhicle was only a quick fix and you already removed the lines (as I intended). Since there is nothing in maximumSafeFollowSpeed that reacts to the 'usage' parameter, you can safely remove it.
2. simply add another commit that reverts this particular change. Then I can squash the PR and it will be gone.

Domsall:
1. `usage` is used in `maximumSafeFollowSpeed` in the same way as `MSGlobals::gComputeLC`?! If I leave out `usage`, tests change. That is why I replaced `MSGlobals::gComputeLC` with `usage`.
2. Ok, I will do this. I will then also update the EIDM-tests.

Domsall:
Done

namdre:
> usage is used in maximumSafeFollowSpeed in the same way as MSGlobals::gComputeLC?! If I leave out usage, tests change. That is why I replaced MSGlobals::gComputeLC with usage.

- As I tried to explain earlier, gComputeLC will remain in the code since it is used in other parts that are not related to carFollowing (i.e. MSLink). 
- also, getMaximumSafeFollowSpeed is not virtual, hence no custom carFollowingModel will need to touch this
- since getMaximumSafeFollowSpeed was working fine with using gComputeLC it can continue to use this and doesn't need to be changed
- the changed CACC test (output/fcd-output/params_caccVehicleMode) is due to setting a parameter conditional on the 'usage' parameter directly in the model. I think this one is ok
- changed IDM test (devices/driverstate/queue_stop_start_idm) is due to supplying the 'pred'  object as new argument in `insertionFollowSpeed` since this influences the driver state
- taking out usage from 'getMaximumSafeFollowSpeed' has no effect on tests

I've added debug code to `getMaximumSafeFollowSpeed` and confirmed that the 'usage' branch is never triggered

```
if (gap < 0 && !MSGlobals::gComputeLC && usage != CalcReason::CURRENT) {
        std::cout << SIMTIME << " debug usage=" << usage << "\n";
    }
```

Domsall:
I had a bug in my Code and just fixed it, sorryfor the confusion!

When you delete `usage` from `MSCFModel::maximumSafeFollowSpeed`, test bugs/8xxx/ticket8817 changes.
If these changes are intended, I will undo the Code parts.

Again, sorry for all the additional work. I thought you added `gComputeLC` in your quick fix not only for the EIDM, but also for followSpeed->maximumSafeFollowSpeed (which also uses `gComputeLC`).

namdre:
I still cannot reproduce the changing behavior in bugs/8xxx/ticket8817 when applying the current diff of this PR and removing/keeping the usage in maximumSafeFollowSpeed. 
(I'm switching between the lines
```
 if (gap >= 0 || MSGlobals::gComputeLC || usage != CalcReason::CURRENT) {
```
and
```
if (gap >= 0 || MSGlobals::gComputeLC) {
```
)

The reason why gComputeLC is used in `maximumSafeFollowSpeed` is because negative gaps have different semantic when considering a leader on the current lane and a leader on a neighboring lane. This is unrelated to the CURRENT and CURRENT_WAIT semantic.  (So please just keep maximumSafeFollowSpeed as it was).

Also, the missing onInsertion parameter bug that you fixed in 6eaf3be is still present in the ACC and CACC models for this PR.

Domsall:
Sorry for the back and forth.
I left the ticket8817-test outdated (which was the reason for the changes). Please update after merging.

namdre:
I was finally able to see the failing test (not sure why I couldn't before). The new behavior is correct and the previous behavior was an unintended side effect of my quick fix (`/// @fixme: setting gComputeLC is workaround until implementing #11265`)

Now I understand why you thought that the new flag was needed in maximumSafeFollowSpeed. ... 
It's fine now (-:

PR #11583, opened 2022-09-13
Refactored function AdaptToJunctionLeader

Domsall:
this should not change any model except EIDM (one EIDM test changed as expected)

### DIFF
--- a/src/microsim/MSVehicle.cpp
+++ b/src/microsim/MSVehicle.cpp
@@ -2982,28 +2982,12 @@ MSVehicle::adaptToJunctionLeader(const std::pair<const MSVehicle*, double> leade
         if (!MSGlobals::gSemiImplicitEulerUpdate) {
             vsafeLeader = -std::numeric_limits<double>::max();
         }
-        if (leaderInfo.second >= 0) {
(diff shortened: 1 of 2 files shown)


namdre:
As far as I understood, the goal of this PR is to eliminate the line `vsafeLeader = MAX2(vsafeLeader, vStop);` since this confuses EIDM.  Why then is it ok for the line `vsafeLeader = MAX2(vsafeLeader, MIN2(v2, vStop));` to remain ?

Domsall:
That is exact the line I had the most problems with...
The workaround was to add CURRENT_WAIT to the stopSpeed-calls. The EIDM will then check in finalizeSpeed, if the stopSpeed-calls were the reason for the next speed update (MAX2 does not matter then). The EIDM still does not know anything about v2, which could theoretically cause some weird behavior. Nevertheless, I could not find any cases, where this was a problem.

In all other cases, the changes work as expected for the EIDM (with CURRENT as input for stopSpeed).

namdre:
Frankly, I find the current code in that function hard to read and this PR isn't helping (by adding some redundancy). I'll try to think of ways to refactor the original to both improve readability and get rid of the MAX calls.

Domsall:
Yes, I had a hard time understanding/changing it all, too. But it works great by comparing, if the vehicle's following speed is sufficient or if the vehicle needs to stop at the junction. 

Thanks in advance!

PR #11591, opened 2022-09-13
Jupedsim integration suggestions

Ozaq:


Ozaq:
@behrisch / @bcoueraud87 please have a look at the changes, I think this might help a bit with the camke integration of jupedsim into sumo.

Be aware that this assumes an installed version of jupedsim.

PR #11648, opened 2022-09-20
Update Emergency.md

DFazekas:


behrisch:
Thanks for that one! I am not a native speaker at all but my gut feeling would be, if I end every item in a bullet list with a full stop it should have a capital letter in the beginning. Do you have a reference style guide you follow?

PR #11699, opened 2022-09-26
Fixed netedit reroute parsing bug

VolkerLieber:
When opening an additionals file, containing a closingreroute with only the dissallowed tag, in netedit, the warning `SVCPermissions must be specified either via 'allow' or 'disallow'. Ignoring 'disallow'` is displayed and the default allow `authority` is used for the specific reroute.
This happens because `AdditionalHandler::parseClosingLaneRerouteAttributes` and `AdditionalHandler::parseClosingRerouteAttributes` always default allow to `authority `if not specified, even when `disallow `is specified by the user.


behrisch:
Hi Volker thanks for the PR. Could you please do the paperwork (open an Eclipse account and answer the ECA)

palvarezlopez:
@VolkerLieber integrated in Main. Thanks!

PR #11702, opened 2022-09-26
Modification of duaiterate.py new 

bmbamdad:
Now separating the pull requests for duaiterate and duraiteratemix


bmbamdad:
- I made another pull request to separate duaIterate and duaIterateMix but it does not work! 
- I rewrite the part of the codes which includes hard coding in duaiterateMix
- I provided an example for the duaIterateMix in a zip file 
- I have already used all of the functionalities of duaiterate in duaIterateMix. But I did not understand what other functionality I should add to duaIterateMix (Let me know if I can improve it by an example please)

namdre:
Thank you for test files. Unfortunately, we cannot use these files for our nightly tests since we have strict time requirements here.
We require example inputs that 
- run to completion quickly (at most 10s)
- use small input files (net ~ 10-100kb, 10-100 trips)

It's enough to attach a zip with test inputs in a comment to this PR.

With regard to refactoring, it should be possible to re-use the initOptions method from duaIterate.py and only add additional options needed for the mix version to the returned options instance

bmbamdad:
Here is a simple example 
[Test.duaIterateMix.zip](https://github.com/eclipse/sumo/files/9683725/Test.duaIterateMix.zip)

Issue #11679, opened 2022-09-23
CACC performs emergency braking at low speed when approaching stopped vehicle


namdre:
- happens at low speeds (regardless of road speed limit)
- doesn't happen at step-length 1s but speed oscillations quickly increase when lowering step length (example runs with 0.1)
- reproducible since at least 1.3.1

namdre:
The model oscillates between these two states

- applying speedControl (previous)
- speedGapControl + applying collision avoidance

The Krauss-based speed override does not come into play and the behavior takes place at speeds below 5m/s

@robertalms Would you care to take a look at this?

robertalms:
Sure, I'll have a look at it.

robertalms:
I can only speculate that nobody noticed this issue so far, since the CACC model was only used in the later half of the TransAID project for a very limited number of use cases (3 that I know of). Those were mostly highway scenarios to investigate traffic flow behavior with ToCs. Probably no one looked into this low speed behavior in more detail. Nevertheless - it's pretty odd

namdre:
some acceleration spikes remain:

![Bildschirmfoto von 2022-09-27 16-34-15](https://user-images.githubusercontent.com/6240630/192555906-5cf3df61-5d94-4d57-ad5f-d8606fb69028.png)

robertalms:
@namdre Your initial guess of adding a spacing error of somewhat around 2m was pretty accurate. I came to a value of > 1.66 . Adding this to the time gap condition in line 365 (time gap > 2) eliminates those remaining spikes for the standstill scenario. 

The micro oscillations you discovered for the platooning scenarios stem from the model frequently switching between mode (ii) and (iv) because the spacing error oscillates around zero... that's part of the original implementation/parametrization of the CACC model and should be tackled in another ticket.

namdre:
Thanks! Would you do a pull request with the proposed changes?

robertalms:
PR done. 

Feel free to change the speed division with an epsilon (NUMERICAL_EPS_SPEED?), as discussed - I didn't really know how to do that. (From what I saw on in the platoon tests, a time gap around 1sec, would be reasonable, even with speed close to zero.)

namdre:
Hm. I'd say it's better than before but there is still something "creepy" about the way the vehicle fails to stop for ~20s and then there is still an acceleration spike at the end:

[...]

PR #11745, opened 2022-09-30
Added spacingError to enable SpeedControlMode in CACC cf_model

robertalms:
@namdre as discussed, here's my pull request for the CACC model to fix standstill/slow approach behavior. You should also consider to make this parameter configurable. 
(Additionally some extensions for the DEBUG output and a name change for the speed error.) ref #11679


namdre:
For future reference: 
Splitting  
- a) the refactoring of variable names and
- b) the fix that introduces a new variable (which happesn to have the same name as the refactored variable)
 
into different commits would have made verifying the change much easier.

PR #11750, opened 2022-10-02
Title attribute to visualisations

### LINKED ISSUE #11646: Feature request: `title` attribute: methods in `tools/visualization`
Feature request:
- Adding a `title` argument to the plotting methods that use matplotlib, by using the following code segment to enhance the plots:
```py
plt.title(given_title)
``` 
- For example, `Linear graph` in
linear_graph
- To maintain compatibility with the existing code, a default title (`None`) will be kept in the argument parsing routine. 

If this seems like a good idea, please assign me to the same.

Thanks,
Vijay

vijay-jaisankar:
closes #11646 

## In this PR, the following features have been added:
- The addition of a `title` attribute to the variable space of plots

## Notes
- In `plotXMLAttr.py`, `sumolib.visualization.helpers` is not used for argument parsing, so the title attribute has been added "manually" by augmenting the parser and calling `plt.title` for the same.


vijay-jaisankar:
It appears that the failing tests are because of `texttest`, which is outside the scope of this PR.

behrisch:
all the scripts which use visualization helpers (plot_csv-...) already have the `--title`-option don't they?

vijay-jaisankar:
> all the scripts which use visualization helpers (plot_csv-...) already have the `--title`-option don't they?

The visualization helpers themselves have this attribute, however the files modified in this PR create the plots with the default value of `title`, as indicated in the helper file (i.e `None`), as the title attribute to be set in the helper functions aren't defined. 

This PR takes the title input from the user and sets it in the helper's namespace.

behrisch:
This won't work. If I add another title option for instance to plot_csv_timeline it conflicts with what is already in helpers.py

PR #11778, opened 2022-10-07
Added dynamic spacing margin to ACC model

### LINKED ISSUE #4551: cfmodel ACC is creating collisions

robertalms:
@namdre: I added the dynamic spacing margin definition from Xiao et. al. 2018 to the ACC model. This fixes #4551, but it might affect other tests as well, since it is a significant change. (The platooning ones I tested looked fine, though.)

Additionally, I suggest to run those ACC tests, also with step-length 0.1 in order to see, if the model is collision free (default seems to be 1s).

A similar spacing margin is part of the CACC model definition - it raises some other issues with the tests, unfortunately and should be handled in a different issue/PR.


robertalms:
I installed TextTest in the meantime. If you give me a quick intro, I could help out reviewing the other tests, as well.

robertalms:
@namdre  I double checked the fcd-outputs for the different tests - they all look reasonable. Below you can compare the accelerations for the discussed use case, as an example. I think those looked definitely improved compared to the old fcd-output.
 
Therefore, I pushed some additional commits to the pull request to fix the compiler error, the updated reference and  updated expected test-results.

plot
plotOLD

PR #11879, opened 2022-10-24
Additional route file support for cycle adaption script

jmlouw:
Currently the `tools/tlsCycleAdaption.py` script only supports `route.rou.xml` files that look like:

```XML
<routes>
    <vType id="type1" accel="0.8" decel="4.5" sigma="0.5" length="5" maxSpeed="70"/>
    <vehicle id="0" type="type1" depart="0" color="1,0,0">
      <route edges="beg middle end rend"/>
    </vehicle>
</routes>
```

However there is the option to specify routes and trips like this as well:

```XML
<routes>
    <vType id="type1" accel="0.8" decel="4.5" sigma="0.5" length="5" maxSpeed="70"/>

    <route id="route0" color="1,1,0" edges="beg middle end rend"/>

    <vehicle id="0" type="type1" route="route0" depart="0" color="1,0,0"/>
    <vehicle id="1" type="type1" route="route0" depart="0" color="0,1,0"/>

</routes>
```

This PR add support for the second type of trips XML file.

### DIFF
--- a/tools/tlsCycleAdaptation.py
+++ b/tools/tlsCycleAdaptation.py
@@ -30,8 +30,12 @@
 - If the critical flow or the sum of the critical flows is larger than 1,
  the optimal cycle length will be set to 120 sec.
 """
-from __future__ import absolute_import
-from __future__ import print_function
+
+# from __future__ import absolute_import
+# from __future__ import print_function
+import logging
+logging.basicConfig()
+logger = logging.getLogger(__name__)
 
 import collections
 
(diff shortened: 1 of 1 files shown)


namdre:
Thanks. However, we can only accept PRs if ECA is in order (https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work)

Also, we are aiming to have our code downwards compatible with python2.7 so please don't just remove any `from __future__ import

jmlouw:
Thanks for the feedback. I signed the ECA - hope that it is order. 

Sorry didn't realise that I never 'uncommented' the `__future__` imports. I added them back again. 

I'd also like to add some `tests`. Where is the best place to do that?

namdre:
ECA looks good and the PR looks good as well.
The place for additional test cases is `sumo/tests/tools/tls/tlsAdaptation`. See https://sumo.dlr.de/docs/Developer/Tests.html
Easiest thing would be to copy an existing test and then tweak the `input_routes.rou.xml` file.
Let me know if you need help with this.

Issue #8529, opened 2021-04-21
Junction "collision" and Turning Angles with lcSigma


Domsall:
I have 2 issues which occur with the sublane model. When simulating solely with sumo, these issues don't really matter, but I'm having problems converting the vehicle trajectories to Carla. I added 2 sumo-configs as examples:

[JunctionAngleTurning.zip](https://github.com/eclipse/sumo/files/6353081/JunctionAngleTurning.zip)

First, the vehicles drive to far onto some junctions and therefore create collisions in an 3D-environment (SUMO real-world viewing). Is this something, that can be dealt with better parameters or junction layout? Or is this something that can/should be changed in the Code?!
And secondly, I simulate with lcSigma > 0 to have some kind of "unperfect lane keeping". This works well, despite the fact that vehicles then don't turn so nicely on the junctions anymore. With lcSigma=0, the vehicle angles get updated continuously, which looks perfect.

Thanks in advance!!!

namdre:
regarding "driving to far onto the junction". This is likely a case of  #4397 and can be fixed manually by moving the internal junction back (connection attribute contPos).

regarding the angles: it would be great if you could prepare a set of small tests that showcase the difference between the current and your proposed angle calculation.

Domsall:
No problem at all, you can't do everything ;-)

what do you have in mind when refering to tests? Short video clips?

I was also thinking of updating the slope calculation similar to the angle calculation (front and back position), because when using the slope for vehicle resistance calculation or when using SUMO in a co-simulation framework, small z-differences can have a huge impact on the slope when points lie close to eachother.

namdre:
patch format is what you get with `git diff src > myPatch`

Domsall:
The Patch:
[AnglePatch.zip](https://github.com/eclipse/sumo/files/6465726/AnglePatch.zip)

And 4 sumocfg-Files including lane changes + junction turns with LC2013/SL2015, each with vehicles types with lcSigma>0 and lcSigma=0.

[JunctionAngleTurning_4sumocfgs.zip](https://github.com/eclipse/sumo/files/6465994/JunctionAngleTurning_4sumocfgs.zip)

namdre:
I think i've found another issue with angle calculation that can only be fixed by considering the history of the maneuver.
Consider a vehicle that turns in a tight circle (i.e. to change lane when close to an obstacle)
- the forward speed will be quite low (maybe 0)
- the lateral speed may also be low in each step
[...]

Issue #5879, opened 2019-08-06
Weird angle visualization during lane-change at junction


schwamborn:
sumo-gui shows weird angle during lane-change at junction:

At step t (bug):

![lc_over_junction_angle_bug_t](https://user-images.githubusercontent.com/35012744/62536347-132a2600-b84e-11e9-85b8-9d0990011ebe.png)

At step t+1 (as it should be):

![lc_over_junction_angle_bug_t+1](https://user-images.githubusercontent.com/35012744/62536353-16bdad00-b84e-11e9-89e1-77275a7244e1.png)

schwamborn:
Added minimal working example scenario to `5879_bad_angle_during_lc` bug report folder.
Occurs at time 17.100.

m-kro:
Probably due to internal lane with length=0.

namdre:
@m-kro can you add a small test to our test suite that captures the changed computation? fcd-output records angles.

PR #11882, opened 2022-10-24
Turning angles with continuous lane changes and lcSigma

Domsall:
This pull request belongs to #8529.

The "new" Code part is the geometric calculation in calcAngleOffset and works as described in the last image.
As it uses speedLat, I needed to make sure that it always gets calculated.

I did not want to update all the tests that now change because of the fcd-output, so please do some tests before integrating.

### DIFF
--- a/src/microsim/MSLaneChanger.cpp
+++ b/src/microsim/MSLaneChanger.cpp
@@ -490,6 +490,7 @@ MSLaneChanger::continueChange(MSVehicle* vehicle, ChangerIt& from) {
     }
     if (!lcm.isChangingLanes()) {
         vehicle->myState.myPosLat = 0;
+        lcm.setSpeedLat(0.0);
         lcm.endLaneChangeManeuver();
     }
     lcm.updateShadowLane();
--- a/src/microsim/MSLaneChanger.cpp
+++ b/src/microsim/MSLaneChanger.cpp
@@ -292,8 +292,9 @@ MSLaneChanger::change() {
 
     if (vehicle->getLaneChangeModel().isChangingLanes() && !vehicle->getLaneChangeModel().alreadyChanged()) {
         return continueChange(vehicle, myCandi);
+    } else if (!(vehicle->getLaneChangeModel().getSigmaLat() > 0)) {
+        vehicle->getLaneChangeModel().setSpeedLat(0);
     }
-    vehicle->getLaneChangeModel().setSpeedLat(0);
     if (!myAllowsChanging || vehicle->getLaneChangeModel().alreadyChanged() || vehicle->isStoppedOnLane()) {
         registerUnchanged(vehicle);
         if (vehicle->isStoppedOnLane()) {
--- a/src/microsim/MSLaneChangerSublane.cpp
+++ b/src/microsim/MSLaneChangerSublane.cpp
@@ -473,6 +473,11 @@ MSLaneChangerSublane::startChangeSublane(MSVehicle* vehicle, ChangerIt& from, do
     const double posLat = -vehicle->myState.myPosLat; // @todo get rid of the '-'
     const double lefthandSign = (MSGlobals::gLefthand ? -1 : 1);
     Position p1 = vehicle->getLane()->geometryPositionAtOffset(vehicle->myState.myPos, lefthandSign * posLat);
+    if (p1 == Position::INVALID && vehicle->getLane()->getShape().length2D() == 0. && vehicle->getLane()->isInternal()) {
+        // workaround: extrapolate the preceding lane shape
+        MSLane* predecessorLane = vehicle->getLane()->getCanonicalPredecessorLane();
+        p1 = predecessorLane->geometryPositionAtOffset(predecessorLane->getLength() + vehicle->myState.myPos, lefthandSign * posLat);
+    }
     Position p2 = vehicle->getBackPosition();
     if (p2 == Position::INVALID) {
         // Handle special case of vehicle's back reaching out of the network
--- a/src/microsim/MSLaneChangerSublane.cpp
+++ b/src/microsim/MSLaneChangerSublane.cpp
@@ -470,7 +470,25 @@ MSLaneChangerSublane::startChangeSublane(MSVehicle* vehicle, ChangerIt& from, do
     // compute new angle of the vehicle from the x- and y-distances travelled within last time step
     // (should happen last because primaryLaneChanged() also triggers angle computation)
     // this part of the angle comes from the orientation of our current lane
-    double laneAngle = vehicle->getLane()->getShape().rotationAtOffset(vehicle->getLane()->interpolateLanePosToGeometryPos(vehicle->getPositionOnLane())) ;
+    const double posLat = -vehicle->myState.myPosLat; // @todo get rid of the '-'
+    const double lefthandSign = (MSGlobals::gLefthand ? -1 : 1);
+    Position p1 = vehicle->getLane()->geometryPositionAtOffset(vehicle->myState.myPos, lefthandSign * posLat);
+    Position p2 = vehicle->getBackPosition();
+    if (p2 == Position::INVALID) {
+        // Handle special case of vehicle's back reaching out of the network
+        if (vehicle->getFurtherLanes().size() > 0) {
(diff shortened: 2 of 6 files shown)


Domsall:
I just recognized that my last commit 

```
if (p1 == Position::INVALID && vehicle->getLane()->getShape().length2D() == 0. && vehicle->getLane()->isInternal()) {
    // workaround: extrapolate the preceding lane shape
    MSLane* predecessorLane = vehicle->getLane()->getCanonicalPredecessorLane();
    p1 = predecessorLane->geometryPositionAtOffset(predecessorLane->getLength() + vehicle->myState.myPos, lefthandSign * posLat);
}
```

does something similar, that is already done a few lines later (was implemented a long time ago: https://github.com/eclipse/sumo/commit/33d7c915187b9d4b6bdac1120b1e304717e0d8bc):

```
if (vehicle->getLane()->getShape().length2D() == 0) {
    if (vehicle->getFurtherLanes().size() == 0) {
        laneAngle = vehicle->getAngle();
    } else {
        laneAngle = vehicle->getFurtherLanes().front()->getShape().rotationAtOffset(-NUMERICAL_EPS);
    }
}
```

At this point, I am not sure which implementation works better.

namdre:
Was your last commit added in response to a test case where my old code didn't work? If so, attach the test case and I'll check.

namdre:
Something else: so far the relationship between lateral position and lateral speed was the same as for longitudinal position and longitudinal speed (for the default non-ballistic update):   `deltaPos = speed * stepLength. `
This means, if the position changes from one step to the next, the speed isn't 0.
With your change in `MSLaneChanger::continueChange` of `lcm.setSpeedLat(0.0);` this no longer holds.

namdre:
I looked at the angle during continuous lane change (lc_model/continuous_lanechange/basic)
Here is the old plot:
Bildschirmfoto von 2022-11-01 10-23-46

Here is the new plot:
Bildschirmfoto von 2022-11-01 10-24-07

I find the new shape surprising.


For the sublane model I looked at test sublane_model/strategic/overtake_stopped_on_the_left

old:
Bildschirmfoto von 2022-11-01 10-29-57

new:

Bildschirmfoto von 2022-11-01 10-30-54

again there is a small plateau before the maneuver ends though it's less pronounced. Any idea whats happening there?

Third question: what tests are good to look at to see the improvement of the new code? Or do we need more tests?

namdre:
FYI: I don't think i'll manage to merge this into the 1.15.0  release but would like to merge post-release

Domsall:
Thanks for your feedback and sorry for the late response, I took a few days off.

> Was your last commit added in response to a test case where my old code didn't work? If so, attach the test case and I'll check.

It corresponds to issue #5879, which has the test sumo\bugs\5xxx\ticket5879. You can change `--lanechange.duration 2.5` to `--lateral-resolution 0.8` and should see a difference, if the edges "begin" and "end" have some curvature.
After looking into it a bit more, I would delete https://github.com/eclipse/sumo/commit/33d7c915187b9d4b6bdac1120b1e304717e0d8bc (old angle calculation from one point) and keep the part from the last commit (angle calculation from two points).
**How can I further help with this?**

> Something else: so far the relationship between lateral position and lateral speed was the same as for longitudinal position and longitudinal speed (for the default non-ballistic update): deltaPos = speed * stepLength.
This means, if the position changes from one step to the next, the speed isn't 0.
With your change in MSLaneChanger::continueChange of lcm.setSpeedLat(0.0); this no longer holds.

You are absolutely right. I thought the same when changing this. The issue I am having is: When the lane change ends, SpeedLat and the Angle get calculated one last time in the LC-part of the Code. In the next step, SpeedLat is still "active" in the Angle calculation, because the Angle gets updated in the CF-part and SpeedLat gets updated after that in the LC-part of the Code. computeAngle() could be added to the LC-part. But then it would always be calculated twice (or we delete it from the CF-part).
**Do you have a good suggestion?**

> I looked at the angle during continuous lane change (lc_model/continuous_lanechange/basic)

I actually used that test as reference myself. The differences we see come from:
1. The lane change has 3 parts: Turning the steering wheel to a specific value, holding it there and turning the wheel back. When the speed and speedLat (which depends on lane coordinates, not vehicle coordinates) stay constant, I expect the angle to have a constant part. If we change the lanechange.duration to higher values, this constant part should become larger and vice versa.
2. The vehicle is still accelerating here. After the acceleration reaches zero, the angle stays constant.

**Or did I missunderstand something here?**

> sublane_model/strategic/overtake_stopped_on_the_left

I do not know what happened the last commits, but my results look different:
image
and
image
This also shows the problem from above, computeAngle() is calculated with the SpeedLat-value from the previous step when the lane change already finished. (this is done here, because `isChangingLanes()` does not exist for the sublane model and therefore `lcm.setSpeedLat(0.0);` is never set.)

> Third question: what tests are good to look at to see the improvement of the new code? Or do we need more tests?

There should be improvements for most of the tests. You could also add the tests posted in #8529.

Independent of the previous questions, I found another issue: `getBackPosition()` does not work correctly with the sublane model, because `isChangingLanes()` seems to only exist for the continuous model.
**Is this true?**

**Please tell me how/where I can help for a fast integration.**

Domsall:
@namdre: a short reminder, if this got stuck in the stack :)

I can also seperate this issue into smaller ones. Anything that helps getting this fixed

namdre:
Hi @Domsall, indeed the issue got buried in the stack and since it has quite a big impact on many tests it's not easily squeezed into my work queue.  If you can separate out different aspects into distinct PRs that would probably help.

(i.e.
- a changeset that only affects the final angle update (if you can put some plots into the PR that show the improvements in before/after angle timelines that would be awesome too)
- a changeset that integrates the new angle computation code (same with the plots)
- A new ticket and discussion for  the getBackPosition issue.

To answer at least an easy question:

> The issue I am having is: When the lane change ends, SpeedLat and the Angle get calculated one last time in the LC-part of the Code. In the next step, SpeedLat is still "active" in the Angle calculation, because the Angle gets updated in the CF-part and SpeedLat gets updated after that in the LC-part of the Code. computeAngle() could be added to the LC-part. But then it would always be calculated twice (or we delete it from the CF-part).

Right now the sublane model does update the angle a second time and I'd prefer it if the continuous model would do the same (with a final calculation once the maneuver is completed).

PS: I'm also aware of #11583 ...

Domsall:
> Right now the sublane model does update the angle a second time and I'd prefer it if the continuous model would do the same (with a final calculation once the maneuver is completed).

From what I understand, the angle already gets updated a second time in both models, but only during the LC-maneuver. When the maneuver finishes, the angle does not get updated in the LC Code part anymore (in both models).
The issue in an example:

1. timestep 0.1s, PosLat 0.05m, SpeedLat 0.5m/s, Angle gets calculated twice
2. timestep 0.2s, PosLat 0.00m, SpeedLat 0.5m/s, Angle gets calculated twice
3. timestep 0.3s, PosLat 0.00m, SpeedLat 0m/s, Angle gets calculated once (and in my case with SpeedLat 0.5m/s, because SpeedLat changes in the LC part of the Code and the Angle gets calculated before that)

The solution from my previous post: computeAngle() could be added to the LC-part. But then it would **always** be calculated twice (or we delete it from the CF-part).

namdre:
I have another idea: we can delay the call to `lcm.setSpeedLat(0.0);`  until the next step so it won't affect the output of the step where the change was concluded.
Just put it into `MSVehicle::computeAngle`:

```
if (!lcm.isChangingLanes()) {     
        lcm.setSpeedLat(0.0);     
 }
result += lefthandSign * myLaneChangeModel->calcAngleOffset()
```
in the next step when the angle is being computed, we can simp

Domsall:
yes, that is a good idea and brings me to the last issue: The sublane-model does not update the variable `myLaneChangeCompletion`, which is used in `isChangingLanes()`.

I will create a new issue for this one.

namdre:
This is because the sublane model can dynamically change the direction and distance while conducting a maneuver. Correspondingly it doesn't use a completion value but only tracks the remaining maneuver distance `myManeuverDist`.

Domsall:
I fully understand this, but how should the previous if-clause look to also work for the sublane model?! What is the sublane counterpart?

namdre:
`getManeuverDist() == 0` should do the trick (though there might be some lose ends where it isn't fully reset yet).

PR #11932, opened 2022-10-31
[tools/randomTrips] Improved the --verbose option.

bcoueraud87:
Note: print statements that are followed by a sys.exit(1) call are still always executed even if the verbose mode is not set, because I think it makes sense.


behrisch:
I think this is true for all error messages

PR #11982, opened 2022-11-08
[tools/plotXMLAttributes] Added the possibility to use wildcards

bcoueraud87:
Concerning randomTrips, I forgot a verbosity check.


namdre:
Please remove the randomTrips changes. The code was fine before because errors and warning must be printed regardless of  option **--verbose**.  (also it's generally better to do these things in different pull requests anyway)

PR #11998, opened 2022-11-09
[tools/plotXMLAttributes] Added wildcards for use in IDs

bcoueraud87:
Now the user can type `--filter-ids bus_*`.


namdre:
For some reason there is still a merge conflict. I tried to resolve it via the git UI but to no avail. Can you try to redo the PR against sumo/main again?

PR #12018, opened 2022-11-10
docs: Instructions for Dockerfile.fedora

vijay-jaisankar:
In this PR, the following functionality has been added:
- Build and run instructions for `Dockerfile.fedora`

The format of these instructions conform to those present in the other Dockerfiles.


behrisch:
maybe we should actually encourage people to name their docker image something like sumo:fedorarawhide here

vijay-jaisankar:
Thanks for the feedback!

I have updated the installation instructions with the new name, please check.

behrisch:
Thank you!

PR #12044, opened 2022-11-15
more helpful error background images message

jmlouw:
osmWebwizard throws an exception when the `satellite background` option is checked. 
The exception looks like:
```
Downloading background images
Error while downloading background images
```
But doesn't elaborate on what went wrong.

This PR throws the exception along with the error message, like:
```
Downloading background images
Error while downloading background images: No module named pyproj
```

In my case the Python module `pyproj` was missing and I had to manually install it. Perhaps this can be part of a requirements file before running osmWebwizard?

### DIFF
--- a/tools/osmWebWizard.py
+++ b/tools/osmWebWizard.py
@@ -293,10 +293,10 @@ class Builder(object):
                 tileGet.get(tileOptions)
                 self.report("Success.")
                 self.decalError = False
-            except Exception:
+            except Exception as e:
                 os.chdir(self.tmp)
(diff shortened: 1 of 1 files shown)


namdre:
Thanks!

namdre:
> Perhaps this can be part of a requirements file before running osmWebwizard?

We don't want to raise hurdles for people that do not need the satellite background. However, we could conditionally disable the 'satellite' checkbox and add a tooltip on the checkbox (@angelobanse  ?)

PR #12045, opened 2022-11-15
bug fix unified cycle doens't have the exact same cycle length

jmlouw:
When the `unicycle` option is enabled when using the `tools/tlsCycleAdaptation.py` script, the cycle times of different traffic lights are not always the same. This bug occurs when the `phase durations` of a traffic light are rounded, the total cycle time can increase or decrease by a few seconds.

A simple example: If the total cycle time is `100` seconds and there are three phases with equal phase lengths, the resultant phase durations will be `33.33` seconds for each phase. If we round this to `33` seconds for each phase, we lose one second from the total cycle time (which is now `99` seconds). This will cause some drift among traffic lights because some will have a `99` second cycle time and others will have a `100` second cycle time, and so on.

This error occurs in two places in the `tools/tlsCycleAdaptation.py` script where rounding is used.

This PR corrects the error by distributing the remaining seconds among the traffic light phases. This is still a work in progress because there are numerous ways to fix this bug.

### DIFF
--- a/tools/tlsCycleAdaptation.py
+++ b/tools/tlsCycleAdaptation.py
@@ -369,19 +369,27 @@ def optimizeGreenTime(tl, groupFlowsMap, phaseLaneIndexMap, currentLength, optio
 
     # adjust the green times if minimal green times are applied for keeping the defined maximal cycle length.
     if minGreenPhasesList and totalLength > options.maxcycle and options.restrict:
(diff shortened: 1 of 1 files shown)


namdre:
Thanks!

During cleanup, I've stumbled across #12068 (so just be aware since you are obviously interested in accurate durations).

PR #12139, opened 2022-11-23
Change where SpeedLat is reset

Domsall:
These changes did not change any tests on my side (as expected).

This is part 1 of 3 of the previous PR #11882

### DIFF
--- a/src/microsim/MSLaneChangerSublane.cpp
+++ b/src/microsim/MSLaneChangerSublane.cpp
@@ -278,7 +278,12 @@ MSLaneChangerSublane::abortLCManeuver(MSVehicle* vehicle) {
 #endif
         outputLCEnded(vehicle, myCandi, myCandi, priorDirection);
     }
-    vehicle->getLaneChangeModel().setSpeedLat(0);
+    if (vehicle->getLaneChangeModel().getSpeedLat() != 0) {
+        vehicle->getLaneChangeModel().setSpeedLat(0);
+        // compute new angle here
+    } else {
+        vehicle->getLaneChangeModel().setSpeedLat(0);
+    }
     vehicle->getLaneChangeModel().setManeuverDist(0.);
     vehicle->getLaneChangeModel().updateTargetLane();
 }
@@ -821,7 +826,12 @@ MSLaneChangerSublane::checkChangeOpposite(
         vehicle->getLaneChangeModel().setOwnState(state);
         return startChangeSublane(vehicle, myCandi, latDist, maneuverDist);
     } else {
-        vehicle->getLaneChangeModel().setSpeedLat(0);
+        if (vehicle->getLaneChangeModel().getSpeedLat() != 0) {
+            vehicle->getLaneChangeModel().setSpeedLat(0);
+            // compute new angle here
+        } else {
+            vehicle->getLaneChangeModel().setSpeedLat(0);
+        }
         return false;
     }
 }
--- a/src/microsim/MSVehicle.cpp
+++ b/src/microsim/MSVehicle.cpp
@@ -4313,6 +4313,9 @@ MSVehicle::executeMove() {
             // check (#2681): Can this be skipped?
             myLaneChangeModel->prepareStep();
         } else {
+            if (MSGlobals::gLaneChangeDuration > 0 && !myLaneChangeModel->isChangingLanes() && !(myLaneChangeModel->getSigmaLat() > 0)) {
+                myLaneChangeModel->setSpeedLat(0);
+            }
 #ifdef DEBUG_ACTIONSTEPS
             if (DEBUG_COND) {
                 std::cout << SIMTIME << " veh '" << getID() << "' skips LCM->prepareStep()." << std::endl;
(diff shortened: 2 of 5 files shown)


namdre:
I'm adding my reasoning for doing things different from your PR as comments into the commit.

src/microsim/MSLaneChangerSublane.cpp
```diff
@@ -278,7 +278,12 @@ MSLaneChangerSublane::abortLCManeuver(MSVehicle* vehicle) {
 #endif
         outputLCEnded(vehicle, myCandi, myCandi, priorDirection);
     }
-    vehicle->getLaneChangeModel().setSpeedLat(0);
```
namdre, line 281: not sure why you would put the same line of `setSpeedLat(0) into both branches of if/else

src/microsim/MSVehicle.cpp
```diff
@@ -4313,6 +4313,9 @@ MSVehicle::executeMove() {
             // check (#2681): Can this be skipped?
             myLaneChangeModel->prepareStep();
         } else {
+            if (MSGlobals::gLaneChangeDuration > 0 && !myLaneChangeModel->isChangingLanes() && !(myLaneChangeModel->getSigmaLat() > 0)) {
```
namdre, line 4316: I don't like leaking thos lcmodel details into MSVehicle. Seem though about putting this only into the alternative branch to  to prepareStep (i'd rather call it unconditionally and have prepareStep do changes on top if needed.

src/microsim/lcmodels/MSLCM_LC2013.cpp
```diff
@@ -1076,6 +1076,9 @@ MSLCM_LC2013::prepareStep() {
             scaledDelta = deltaPosLat * myVehicle.getSpeed() / myVehicle.getLane()->getSpeedLimit();
         }
         myVehicle.setLateralPositionOnLane(oldPosLat + scaledDelta);
+        myVehicle.getLaneChangeModel().setSpeedLat(DIST2SPEED(scaledDelta));
+    } else if (!isChangingLanes()) {
+        myVehicle.getLaneChangeModel().setSpeedLat(0);
```
namdre, line 1081: should apply also to DK2008 (and other models people come up with, thus better placed in the base class.

src/microsim/MSLaneChangerSublane.cpp
Domsall: The idea is that the Angle does not always have to be computed twice then.
For better readability it could also look like this:
```
double speedLat = vehicle->getLaneChangeModel().getSpeedLat();
vehicle->getLaneChangeModel().setSpeedLat(0);
if (speedLat != 0) {
    // compute new angle here
}
```

Two issues now: 

- the angle will always be computed twice
- The changes at line 824ff. are not in the commit

src/microsim/lcmodels/MSLCM_LC2013.cpp
Domsall: Does DK2008 have continuous lane changes and/or speedLat?

Why I put `setSpeedLat` in this if/else if clause:

- Each time step, only one `setSpeedLat`-call should be made, because each call updates myAccelerationLat. When simulating with sigma now, myAccelerationLat will not be correct.

src/microsim/MSVehicle.cpp
Domsall: The new code is missing `!(myLaneChangeModel->getSigmaLat() > 0)`. As I wrote in the other comment, `setSpeedLat` should only be called once. The new function is a bit cleaner and seems to (almost) do the same as the Code here and below.

src/microsim/MSLaneChangerSublane.cpp
namdre: I see your point about double computation but I still don't undertand the `else` branch  since it should be a no-op
```
if (vehicle->getLaneChangeModel().getSpeedLat() != 0) {
   ...
    } else {
        vehicle->getLaneChangeModel().setSpeedLat(0);
    }
}
```

Domsall: The problem of the old code was that if we call `setSpeedLat(0)` first and then call `getSpeedLat()`, SpeedLat will always be 0. That is why I first checked if SpeedLat is 0 and then changed it afterwords (but before the angle computation).

src/microsim/MSVehicle.cpp
namdre: My idea was to have the sigmaLat stuff be handled *afterwards* in `prepare step` (so we don't have to worry about the details here). `prepareStep` computes a completely new speedLat anyway when using sigmaLat.

src/microsim/lcmodels/MSLCM_LC2013.cpp
namdre: The continous lane change model is a layer above the individual lane change models so in principle it should work (even with SL2015 when not running with lateral resolution). However, it didn't until now #12144.
The point remains that the feature is not intended to be specific to LC2013.

Good point regarding `myAccelerationLat though`. I'll have to reconsider.

namdre: Actually, we already had a similar problem with `sigmaLat` before: prepareStep updates speedLat and in the same step we may start a new lane change that updates speedLat again.

Issue #12132, opened 2022-11-23
Is it safe to store additional metadata in the parameters map?


dariuszparys:
We're currently using netconvert to convert openstreetmap data into opendrive data. We need to extend the current conversion with additional tags that can be mapped to opendrive data, like e.g. roadMarks. The question I have, is it safe to place additional data into the `NBEdge` parameters map so we can adjust the importer for openstreetmap and exporter for opendrive to leverage the data their?

m-kro:
We already use parameters to store the original OpenStreetMap IDs in the SUMO network, see netconvert option **--output.original-names**. I think there is no problem in reading parameters in your customised exporter.

behrisch:
@dariuszparys And please consider doing a pull request with your enhancements

namdre:
Note that there is already some infrastructure in place to import additional attributes by setting option **--osm.all-attributes** an **--osm.extra-attributes all**

dariuszparys:
Thanks @namdre I indeed have overlooked this. This reduces the changes to the opendrive export.

namdre:
also: roadMarks should already be exported based on lane change prohibitions. If this doesn't work pleas file a bug report.

dariuszparys:
So there is full support for 6.2 Road Mark Type Information from OpenDrive 1.4?

namdre:
at the moment only the marktypes broken,solid and none are used and infered from the lane restrictions on the lane. The lane change restrictions that are already imported into the network model and which could be used to infer "solid broken" and "broken solid" (and which are already painted in sumo gui) are not used in the xodr export. Pull requests are very welcome.

PR #12158, opened 2022-11-25
feature: export osm divider tag to opendrive 1.4

dariuszparys:
This PR allow the export of the experimenental divider tag from OpenStreetMap see https://wiki.openstreetmap.org/wiki/Key:divider

The export converts the following from OpenStreetMap to OpenDrive 1.4

| OpenStreetMap | OpenDrive 1.4 |
|-----|---------------|
| solid_line | solid |
| dashed_line | broken |
| double_solid_line | solid solid |

More tags might be added but that would be a first good start.

## Background

There is a need in our engagement to have the center lane road marks to be of the type specified in the OpenStreetMap definition and exported accordingly in the OpenDrive 1.4 output file. As netconvert is used for purely the conversion without specifying lane change restrictions the option that seems obvious to use is the additional OSM attributes (as asked in #12132) and having a switch to flag specific export behavior and respect divider tag attributes in the conversion.

## Tests

Tests have been adjusted and pass successfully using `texttest`.

## Additional Notes

I'm new to the code base and still try to wrap my head around it.

### DIFF
--- a/data/xsd/netconvertConfiguration.xsd
+++ b/data/xsd/netconvertConfiguration.xsd
@@ -80,6 +80,7 @@
             <xsd:element name="railway.topology.output" type="fileOptionType" minOccurs="0"/>
             <xsd:element name="polygon-output" type="fileOptionType" minOccurs="0"/>
             <xsd:element name="opendrive-output.straight-threshold" type="floatOptionType" minOccurs="0"/>
+            <xsd:element name="opendrive-output.osm-divider-tag" type="boolOptionType" minOccurs="0"/>
         </xsd:all>
     </xsd:complexType>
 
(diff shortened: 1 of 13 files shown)


namdre:
Thanks! I'll be happy to merge the code but I'm still wondering whether this even needs a new option (rather than making it normal behavior).  It's already contingent on importing this OSM tag (which isn't default behavior) and people could chose to not import this tag (via an **--osm.extra-attributes** list).

dariuszparys:
@namdre @behrisch I was not sure if making this the normal behavior would interfere with any other use cases that I might not think of. I can of course remove the switch, let me know.

behrisch:
I am also  fine to have this without the option

dariuszparys:
I removed with the latest commit the switch.

namdre:
Thanks!

PR #12162, opened 2022-11-25
OpenStreetMap width:lanes handling and OpenDrive width computation fix

danigian:
This PR includes changes related to the following two points. For any reviewer, please let me know if you prefer the changes split in two different PRs.

## Importing width:lanes from OpenStreetMap

A new feature that allows to import the [width:lanes](https://wiki.openstreetmap.org/wiki/Lanes) key/subkey combination has been added.
As per OpenStreetMap wiki, the value of the width:lanes tag contains the values for each lane separated by a | (vertical bar) in left-to-right order as viewed in the respective driving direction of those lanes.
Import is taking into account LHT/RHT differences when it comes to lane index handling in SUMO.

A test has been added to cover this new feature.

## Fix: OpenDrive width computation for connecting roads

When exporting to OpenDrive, a constant lane width was used for junctions and connecting lanes.
This resulted in improper conversion for lanes connecting roads/lanes with different width, as can be observed from the rendering of one of the existing test cases for OpenDrive export:

image

OpenDrive is defining the lane width as a third order polynomial in function of the lane distance from its starting point. An accurate definition of the four coefficients of this third order function is only possible if we know at least four points. Unfortunately when it comes to connecting two roads, only the width of the predecessor and the width of the successor is known, therefore we can at the current time only provide a linear function.

This is anyways enough for fixing most of the cases, as can be observed from the rendering of the same test case analyzed before:

image

### DIFF
--- a/src/netimport/NIImporter_OpenStreetMap.cpp
+++ b/src/netimport/NIImporter_OpenStreetMap.cpp
@@ -598,7 +598,8 @@ NIImporter_OpenStreetMap::insertEdge(Edge* e, int index, NBNode* from, NBNode* t
 
     const std::string origID = OptionsCont::getOptions().getBool("output.original-names") ? toString(e->id) : "";
     if (ok) {
-        const int offsetFactor = OptionsCont::getOptions().getBool("lefthand") ? -1 : 1;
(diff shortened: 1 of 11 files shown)


namdre:
Excellent!

Issue #12146, opened 2022-11-24
getBackPosition() uses LC-model-dependent function


Domsall:
In `getBackPosition()`, the function `isChangingLanes()` is called multiple times.
`computeAngle()` uses `getBackPosition()` to calculate the correct position/angle values during continuous lane changes.

I would like to use `getBackPosition()` to do a similar angle-calculation for the SL2015 model (angle between front/back position). 

Right now, this angle calculation does not work due to `isChangingLanes()`/`myLaneChangeCompletion` not being set in the SL2015 model. For the current sublane angle calculation, this is not an issue, because during lane changes the angle-value from `computeAngle()` gets overwritten in `MSLaneChangerSublane::startChangeSublane`.

See #11882 for reference. Part 2 of 3.

namdre:
To clarify: 
- Is the problem limited to situations where `MSLaneChangerSublane::startChangeSublane` is not being called?
- could the problem be solved (less inefficiently) if `MSLaneChangerSublane` would always recompute the angle?
- where exactly is the problem in `getBackPosition`, More specifically, which of the if/else branches is returning a wrong back position and why?  (The branching is mostly related to whether we can make use of furtherLanes or not and this ought to work already in the sublane case)

Domsall:
- I would like to change computeAngle() in a way that it works for all cases in the sublane model (when `MSLaneChangerSublane::startChangeSublane` is called and when not). We then have one implementation and just need to call `computeAngle()` in every case. But yes, `isChangingLanes()` is only erroneous when a lane change is actually occuring (so in `MSLaneChangerSublane::startChangeSublane`).
- No, it is working now because `getBackPosition()` is not yet used in `MSLaneChangerSublane`. Nevertheless, it is needed in the future, because the current implementation is causing issues with SigmaLat (see tests in #8529).
- To see the issue in the current code base: Negate the `isChangingLanes()` calls in `getBackPosition()` and run the test sumo\bugs\5xxx\ticket5879 with the continuous lane change model.

namdre:
Sorry but I'm still unclear on this so let me rephrase my question.

Q2. 
- I think I understood the motivation. These questions are really just for understanding the solution space for implementing a fix.
- I'm aware that problem only comes up once the new angle computation makes use of the back position
[...]

PR #12201, opened 2022-12-01
Add new calcAngleOffset() with known isChangingLanes()-bug

Domsall:
For this to work correctly:

1. ~~`isChangingLanes()` in `computeAngle()` needs to be changed.~~
--> Done, does not need to be changed. I checked, if this Code part changes any test and it does not.
2. ~~`isChangingLanes()` in `getBackPosition()` needs to be changed. #12146~~
--> Done
3. Update tests

See #11882 for reference. Part 3 of 3.

### DIFF
--- a/src/microsim/MSLaneChangerSublane.cpp
+++ b/src/microsim/MSLaneChangerSublane.cpp
@@ -284,7 +284,7 @@ MSLaneChangerSublane::abortLCManeuver(MSVehicle* vehicle) {
     vehicle->getLaneChangeModel().updateTargetLane();
     if (updatedSpeedLat) {
         // update angle after having reset lateral speed
-        //vehicle->computeAngle();
+        vehicle->setAngle(vehicle->computeAngle());
     }
 }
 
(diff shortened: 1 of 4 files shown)


namdre:
Some updates:
- getBackPos should now return sensible values throughout (#12146) - let me know if you still have a test where getBackPos returns bad things)
- I'm okay with having `myPreviousAngleOffset` in the code (the new `computeAngle` will actually make it easier to refactor this out. The old sublane angle computation makes it infeasible to move `computeAngle` to a point in the code where positional updates have been completed).

I'm looking at angle over time  in `lc_model/continuous_lanechange/basic` and testing this with steplengt h1 and 0.2: (green is the old behavior):

step1
step0 2

My assumption would be that the curve keeps some kind of symmetry with respect to the midpoint of the maneuver.

The next plots are from `sublane_model/speedGain/default` with steplength 0.2 (plots are for the ego vehicle overtaking on the left and the non the right.  (**note: the blue is the old behavior here**)

figure_1

Again the new plots are less symmetrical (especially the second maneuver).  The common feature for both tests seems to be that the new code produces a smooth start of the maneuver but a rather sharp angular change at the end of the maneuver.  The smoothness looks like an improvement but the asymmetry makes me wonder whether there is still a bug somewhere.
To rule out that this is an artifact of the sublane model I also looked at lateral speed. It doesn't appear that the lateral speeds have an asymmetry around the midpoint of the maneuver:

speedLat0 2




I also looked at the test lcSigma case you provided in #8529 and the angle plots look indeed much better than before:

sigma

As a side not: The junction collision reported the #8529 test (when activating tracking) is rather due to an invalid internal junction position that can be fixed be saving the network in netedit again. Though I can image that the angle "flips" were causing all kinds of problems in other context.

Domsall:
Thank you very much for extensively testing the new Code!

The first issue with the continuous lane change model comes from the way speedLat is computed.
At the beginning of the lane change, speedLat is set to a specific value. This value stays constant until the lane change ends. The angle does not start changing back to its old value until the lane change is over and speedLat goes from the value back to zero.
2 solutions to change this behavior:

- change the lane change trajectory to something smoother, this would make speedLat change its value smoother (something similar to https://github.com/octavdragoi/sumo/tree/smooth-lane-change)
- add a "hack" that reduces the speedLat-value for the angle-calculation depending on the lane change completion value.

The asymmetry of the angle is a bit expected.
What happens now:

- When the lane change is over (latDist=0), the angle is already back to its old value.

Whats happens with the new Code, if I compare the first angle and last angle of a lane change:

image

It takes a bit longer to "straighten" out. I hope this makes sense.

namdre:
If you look at the second picture,  (lc_model/continuous_lanechange/basic with steplength 0.2) you can see that the weirdness is not something that happens in the last step of the lane change (the final part is nice and smooth). Instead it's somewhere in the "middle" of the maneuver where the angle starts to reverse direction.  So frankly, I still don't understand the shape of the curve.

It's not really worse than the old behavior thouch and the lcSigma behavior is much better. So I don't really mind merging this. 

any comments from your side @behrisch ?

behrisch:
No, just go ahead

PR #12229, opened 2022-12-05
Exporting crosswalk objects to OpenDrive

danigian:
This PR includes changes related to exporting crosswalk objects from SUMO networks to OpenDrive format.

For the reviewer: please give your opinion on changes introduced in "writeRoadObjectPoly" method with regards to the t offset, I had to do the changes for properly handling offsets in lefthand conversions, but still some inaccuracies might be there.

image


namdre:
Thanks! 
I don't have any tests that show a problem with your crosswalk offset code so I'm fine with this until further notice.

PR #12231, opened 2022-12-06
Feature Traffic Signs for OpenDrive

dariuszparys:
This PR modifies the OpenDrive export codebase to support the OpenStreetMap `traffic_sign` tag. This PR also modifies the XML for PolyConvert, I wasn't sure if this should be a separate PR, but the changes are so minimal, I just kept it in this one.

## Conversion of `traffic_sign` to OpenDrive 1.4

Traffic signs will be converted to two elements for the `road` parent in OpenDrive: `object` and `signal`. The `object` tag will be of type `pole` while the `signal` will be translated to a corresponding tag containing `country`, `type`, `subtype` and `value`.  The following snippet should make this clearer.  

The OpenStreetMap declaration  
  
``` xml
<node id='-137713' action='modify' visible='true' lat='48.89047121967' lon='9.1823734761'>
  <tag k='Maxspeed' v='80' />
  <tag k='traffic_sign' v='Maxspeed' />
</node>
```

translates to this shape representation using `polyconvert`  

``` xml
<poi id="-137713" type="traffic_sign" color="black" layer="5.000000" lon="9.182373" lat="48.890471">
  <param key="Maxspeed" value="80"/>
  <param key="traffic_sign" value="Maxspeed"/>
</poi>
```

which results now in the OpenDrive 1.4 output

``` xml
<road>
  <!-- shortened for clarity -->
  <objects>
   <object id="-137713" type="pole" name="pole" s="36.59856896" t="0.30000000" hdg="0"/>
  </objects>
  <signals>
   <signal id="-137713" s="36.60" t="0.30" orientation="-" dynamic="no" zOffset="3" country="OpenDrive" type="maxspeed" subtype="" value="80" height="0.78" width="0.78"/>
  </signals>
</road>
```

The conversion of the `signal` tag is handled in `parseTrafficSign` a function that is introduced with this PR.

## Tests  

Three tests have been added and can be found in the directories  

- osm_traffic_sign
- osm_traffic_signs
- osm_traffic_sign_lefthand
- osm_traffic_sign_LHLL

## Known Limitations

- The conversion is looking for specific `traffic_sign` elements but can be extended. Currently `Maxspeed`, `stop`, `hazard` and `give_way` are supported
- The conversion also supports a basic traffic sign id conversion as defined in [[Key:traffic_sign - OpenStreetMap Wiki](https://wiki.openstreetmap.org/wiki/Key:traffic_sign)](https://wiki.openstreetmap.org/wiki/Key:traffic_sign). Currently basic patterns are supported through a regular expression, like `DE:260,1020-30;265[3.8]`, here a rendering example

multiple-traffic-signs


namdre:
The value of `true` in https://github.com/eclipse/sumo/blob/b8997887d3433f605530eef0f1870a2d895e7252/src/netwrite/NWWriter_OpenDrive.cpp#L1268 ensures that only objects that are at the side of a road (rather than in front or behind) are matched.  All the test cases you added still work with this setting. If you have other cases where this fails please open up a new ticket and attach the test case (or better, add a test case with a PR).

namdre:
@behrisch I'm unsure on whether  traffic signs as pois are a good default when importing OSM (polyconvert),  (see https://github.com/eclipse/sumo/pull/12231/commits/d2e5182387b877782715fd82e44a70aaec2adf04#diff-83c27967fe14003b09025f925cc08a399868ec2aabd21ea6b45751de9e16b9f4R3)

We could add 'discard="true"` and thereby still make it easier for users to discover this feature while keeping it out of the default export.

namdre:
@dariuszparys One more thing: When you merge main into your fork, please do a `pull --rebase` instead of a `pull` since this avoids the need for extra merge commits such as 688e820a866e0e37ebaf1a03d4d433806f6bdbe3 and results in a cleaner commit history.

dariuszparys:
> @dariuszparys One more thing: When you merge main into your fork, please do a `pull --rebase` instead of a `pull` since this avoids the need for extra merge commits such as [688e820](https://github.com/eclipse/sumo/commit/688e820a866e0e37ebaf1a03d4d433806f6bdbe3) and results in a cleaner commit history.

Thanks, haven't thought of this, sorry for any inconvenience

PR #12238, opened 2022-12-06
Import highway=crossing tags from OSM into SUMO

danigian:
## Description

This PR includes changes related to importing crossings from OSM to SUMO networks.

## What changed:

- [x] a new `osm.crossings` option has been added for importing the highway=crossing tag from OSM
- [x] documentation has been adjusted
- [x] a test has been introduced covering the new scenario


namdre:
Something else: The code after https://github.com/eclipse/sumo/blob/887c277c3431a93631c426283c44e17e38e989f7/src/netimport/NIImporter_OpenStreetMap.cpp#L258 has two disadvantages:

- it is incompatible with guessed sidewalks (because those are added only after the OSM importer has finished)
- it has a lot of redundancy with https://github.com/eclipse/sumo/blob/887c277c3431a93631c426283c44e17e38e989f7/src/netbuild/NBNode.cpp#L2565

A better solution could be to store a list of node ids that shall receive crossings (i.e. in NBNodeCont) and trigger guessing for this particular list.

PR #12255, opened 2022-12-08
[docs] Improved the SUMOlympics tutorial

bcoueraud87:


m-kro:
Could you please unify the way you write netedit on this page? I spotted "NetEdit"...

bcoueraud87:
@m-kro I don't see any in the new version, they are all written as netedit.

m-kro:
Here: "We created the vehicle types and the vehicle flows "by hand", by writing the definitions down into a XML file, however these manipulations can directly be made (and visualized) in **NetEdit**!"

bcoueraud87:
@m-kro You're right! I will make the change.

bcoueraud87:
@m-kro Done!

Issue #12140, opened 2022-11-23
allow trains to keep their old rotation on reversal


namdre:
It's a tradeoff between realism and easily seeing the direction of movement.

- new gui setting "keep orientation on reversal"
- all vehicles store their orientation (intialized to forward=true)
- orientation is toggled on every reversal
- rendering uses the 'forward' attribute (for shapes and images)

PR #12314, opened 2022-12-16
 Added guiShape aircraft and support for reversing aircraft/train 

theDivj:
For "real-world" visualisation  - expected to be used with vClass rail and carriageLength=0
  train reversal code is in GUIVehicle.cpp - aircraft shape/reversal only needs the other changes.

Introduction of guiWidth needed to force loading of passengers / containers to be in centre of image - ie as far as the 'model' is concerned width is used as usual - guiWidth is only applied in drawing the aircraft polygon.

The polygon is large and inefficient - however it is difficult to envisage scenarios where the number of aircraft being 'visualised' would ever be high enough for this to matter.

Aircraft polygon retrieved from https://iconsplace.com/wp-content/uploads/_icons/000000/256/png/boeing-767-icon-256.png - image converted to .plt format via convertio.com, then scaled to +/- 1.0 in excel and transposed to row for use in code.


namdre:
I think physical width is a fundamental value and shouldn't be changed just to reflect seating arrangement. I replaced `guiWidth` with `seatingWidth` to reflect this.

namdre:
Sometimes it's useful to visually indicate the forward direction of trains (i.e. the end where the train conductor is sitting). 
While this is currently indicated by the locomotive front window, it's less obvious than the locomotive shading and also doesn't work in triangle or box mode.
Hence I made the reversal optional (#12140).

theDivj:
Many thanks,
That makes much more sense/is much cleaner - I just did not understand enough about exactly how the transportables were positioned to do that.

Reflecting on my thoughts about there never being a large number of aircraft to model: It did occur to me that airways might be able to be modelled with lane changing - ie different heights being modelled with different lanes - you then have Sumo modelling global transport, today Hokkaido, tomorrow the world .... :)

( In working up my scenario around Liverpool airport I noticed a minor visualisation defect: With opposite direction driving, any 'real world' carriage vehicles (trailer, bus/flexible etc) "rotate" when overtaking. I don't think this has been flagged to date so it's not upsetting anybody - merely entertaining! )

cheers
Div

------- Original Message -------

namdre:
Thanks for letting me know! (#12331).

When looking at your scenario, I noticed that sometimes airplanes were going in reverse after turning a sharp corner. 
In such cases it appears more likely they would have done a 3-way turning maneuver and continued in forward direction. You can model this in sumo by adding a route that enters (reverses into) the other direction of a road and then doing an immediate reversal. The only requirement for this maneuver is a turn-around connection on the reversal edge (but the vehicle won't actually drive to the end of that edge).

PR #12353, opened 2022-12-27
correct lane access when importing from OSM

tinova-1:


behrisch:
Thanks a lot for this one! Did you sign the ECA as mentioned here: https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work ? Also please try to separate the code changes from the typo changes. So could you maybe resubmit with only the changes to the OSM importer?

PR #12376, opened 2023-01-04
[JuPedSim] Maintenance commit.

bcoueraud87:


bcoueraud87:
This is essentially a dummy commit that contains dummy code for pedestrian routing as well as dummy code to temporarily take into account a change in JuPedSim's API.

PR #12402, opened 2023-01-06
correct lane access when importing from OSM

tinova-1:
@behrisch This are only the code changes, not the comment typo fixes


behrisch:
Thank you! Are the netgen tests expected to fail? Did you have a look at the ECA?

tinova-1:
Yep, I did register for Eclipse and accepted the ECA. I even added a signoff to my commit. Some tests fail in the commit that I forked from. I was pretty sure that my edits don't fail any tests that succeed in the previous commit (4ce12b2a5d). But I'll check again.

namdre:
@tinova-1 The eclipse website states that `Anonymous (webmaster@****.ch)` cannot be validated whereas `(timo@****.de)` is covered by the ECA. It seems as if some of your commits still have the wrong email.  You could try to copy all your changes into a fresh fork and do a single squashed PR with the validated email address.

tinova-1:
I'm setting up a new fork anyway To Clean Up the Mess

behrisch:
Thank you!

PR #12436, opened 2023-01-11
NetEdit gui change 

mfsen:
there is a strange methodology at changing the super modes because i coundt find the event that shıows which button is pressed.


behrisch:
Thanks for this one! Please try to avoid whitespace changes in existing code and make sure you have the ECA signed, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work

PR #12457, opened 2023-01-13
Fix osm lane access

tinova-1:


tinova-1:
@behrisch It looks like the windows build fails, is there anything I can do about it? Is it my fault at all? Thanks for your help.

behrisch:
Thanks a lot!

PR #12834, opened 2023-03-15
gtfs2pt improvements

gladysmcgan:
**1. Filter stops by most common shape instead of longest shape**
- This filters out stops that are not commonly in route nor likely to be present in OSM route

**2. Map OSM routes without shapes file** 
- Used the same principle as with using shapes, took the first and last stop (rather than first and last shape coordinate) to get an estimate of the direction of route 

**3. Map GTFS to OSM routes only when,**
- they share at least one common stop. This prevents mapping to completely different lines with the same name.
- they have smaller than 160 degrees angle difference. This prevents mapping to route in opposite direction if correct direction is not available.

**4. Parse PT line color from ptlines.xml to create colored vehicles**

- [ ] Requires modifying ptlines.xml prior to add default color, so PT lines without color are also parsed

### DIFF
--- a/tools/import/gtfs/gtfs2osm.py
+++ b/tools/import/gtfs/gtfs2osm.py
@@ -557,7 +557,7 @@ def map_gtfs_osm(options, net, osm_routes, gtfs_data, shapes, shapes_dict, filte
                 map_routes[row.shape_id] = (osm_id, edges.split())
             else:
                 # no osm route found, do not map stops of route
-                missing_lines.append((pt_line_name, sumolib.xml.quoteattr(str(row.trip_headsign), True)))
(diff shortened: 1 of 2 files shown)


behrisch:
Thanks a lot!

behrisch:
I am sorry, I needed to revert it because our tests were failing which I did not notice on the first run. Now the changes are back again except for the trolleybus. Can you maybe send a small example here which we can add to our tests. I don't think we always want to import the whole of Zürich. Also please check whether my refactoring of your changes broke things. Thanks again for the submission

gladysmcgan:
Hi, no worries at all. Here's a small area of Zürich: [zurich_gtfs.zip](https://github.com/eclipse/sumo/files/11200353/zurich_gtfs.zip)

PR #12895, opened 2023-03-23
Remove SonarCloud cache setup as it is now offered by default

mpaladin:
No need to configure the cache anymore, SonarCloud now has an automatic analysis caching. See https://docs.sonarcloud.io/advanced-setup/languages/c-c-objective-c/#analysis-cache.

Also, the cache was not properly set up with the cache action.


behrisch:
Thanks for this! I am willing to give it a try, unfortunately I cannot accept the PR because the paperwork seems to be missing. Did you accept the ECA? https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work

PR #12954, opened 2023-03-30
fix: code blocks

jalemann:
fix: code indentation
fix: missing )
fix: use spaces instead of non-breakable spaces
add: python flag to markdown code blocks containing python code


jalemann:
I found non-breakable spaces in many other files, if you want to stick to them, I can also create a new PR where just indentation, code block flag and missing `)` are fixed, just let me know.

behrisch:
Thanks! I don't think the non breaking spaces have any benefit but maybe you can still separate the two PRs

PR #12994, opened 2023-04-06
Some refactored argument parsers for tools/route

robertalms:
@behrisch As discussed, here some, but not all refactored parsers for the tools/route directory.


m-kro:
You could even remove such argument to ArgumentParser as this is the default if args is not supplied:
`args=sys.argv[1:]`

PR #13004, opened 2023-04-10
Translation to Italian

m-kro:
Mostly GUI, still missing a lot of strings... Hopefully most terms are almost right.


behrisch:
I needed to remove some strings in https://github.com/eclipse/sumo/commit/c93da6fc23290d8e600ea885c516c9d638218f0d which were not UTF-8 encoded. Please redo if possible. Thanks!

m-kro:
I had added those strings using the Weblate web form. Was not aware before that the text is not transmitted in UTF. Now I created bookmarks of special Unicode characters (accents) in Weblate for easy use.

Issue #10901, opened 2022-06-10
retrieve road angle


namdre:
- lane.getAngle(laneID, pos = -1)
 - edge.getAngle(edgeID, pos = -1)

If a negative position is given, the total angle (from start to end) is returned, otherwise the angle at the given position (the value should match the returned angle for a vehicle at that position)

namdre:
The current work-around is to retrieve the lane shape and then compute the angle manually (i.e. with the help of sumolib.geomhelper)

behrisch:
@bcoueraud87 for you!

bcoueraud87:
@behrisch yes, please assign me

behrisch:
I would rather add this to sumolib instead. I think if we start making static info like this available via traci people will (rightfully) expect other things like getLength etc. I do not want to follow that path @namdre Opinions?

namdre:
Maybe getLength is a bad example because we already provide this for traci.lane. Obviously, we need to draw the line somewhere but I think we might just put getAngle on the traci side of the line (seeing how we don't even need a new message id).

There is also the point of how to support users that are not using python.

behrisch:
But getLength is not there for the edge, so maybe it is actually a good example ;-) where we put in some kind of effort but did not have everything the user might expect. I agree on the part about the non-Python users though. So let's just add getAngle and discuss "the line" in the next developer meeting.

behrisch:
@namdre Any objection against using INVALID_DOUBLE_VALUE instead of -1?

namdre:
using INVALID_DOUBLE_VALUE in the output is preferred but it's cumbersome to use in the input whereas -1 is fine as an invalid offset. I guess we could support counting backwards from the lane end with negative values.

Go ahead an use INVALID_DOUBLE_VALUE

namdre:
@bcoueraud87 can you extend the docs at
https://sumo.dlr.de/docs/TraCI/Lane_Value_Retrieval.html
https://sumo.dlr.de/docs/TraCI/Edge_Value_Retrieval.html
?

PR #13159, opened 2023-04-26
Added the getAngle function to TraCI (#10901).

bcoueraud87:


behrisch:
you should add a test for the edge as well but it may be enough to do this in python (complex/traci)

behrisch:
Is there any reason why you switched from libsumo::INVALID_DOUBLE_VALUE back to -1

behrisch:
The traci/get_variable/Lane test still uses -1. Maybe just test with an explicit value here. Also the doc string in _lane.py is wrong. The result is not the angle from the current position to the end. It is the inclination of the straight line segment at the given position (except for the default case where it is from begin to end).

namdre:
I'd rather say it's the orientation because inclination is more suggestive of the z axis

bcoueraud87:
> The traci/get_variable/Lane test still uses -1. Maybe just test with an explicit value here. Also the doc string in _lane.py is wrong. The result is not the angle from the current position to the end. It is the inclination of the straight line segment at the given position (except for the default case where it is from begin to end).

Thanks I will make the changes and sorry for the mistakes.

behrisch:
Another point: SUMO allows for special district edges without lanes. In this case we should return INVALID_DOUBLE_VALUE

bcoueraud87:
> Another point: SUMO allows for special district edges without lanes. In this case we should return INVALID_DOUBLE_VALUE

I will do that as well then.

src/libsumo/Edge.cpp
```diff
@@ -281,6 +282,14 @@ Edge::getPendingVehicles(const std::string& edgeID) {
     return vehIDs;
 }
 
+
+double
+Edge::getAngle(const std::string& edgeID, double relativePosition) {
+    MSLane* lane = getEdge(edgeID)->getLanes()[0];
+    return lane ? Lane::getAngle(lane->getID()) : libsumo::INVALID_DOUBLE_VALUE;
```
behrisch, line 289: This is not correct. If the edge has no lanes, the vector is empty

PR #13180, opened 2023-05-03
Tesselation of polygons with holes.

bcoueraud87:
I can provide coordinates of a polygon with holes if needed. I have tested this code in relation with the JuPedSim coupling.

### DIFF
--- a/src/utils/gui/globjects/GUIPolygon.cpp
+++ b/src/utils/gui/globjects/GUIPolygon.cpp
@@ -39,7 +39,6 @@
 #define CALLBACK
 #endif
 
-
 // ===========================================================================
 // static members
 // ===========================================================================
@@ -119,7 +118,12 @@ TesselatedPolygon::drawTesselation(const PositionVector& shape) const {
     if (myTesselation.empty()) {
         myCurrentTesselated = this;
         // draw the tesselated shape
-        double* points = new double[shape.size() * 3];
+        size_t shapeSize = shape.size() * 3;
+        size_t holesSize = 0;
+        for (PositionVector hole : myHoles) {
+            holesSize += hole.size() * 3;
+        }
+        double* points = new double[shapeSize + holesSize];
         GLUtesselator* tobj = gluNewTess();
 #ifdef _MSC_VER
 #pragma warning(push)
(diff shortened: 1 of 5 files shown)


src/utils/gui/globjects/GUIPolygon.cpp
```diff
@@ -119,7 +118,12 @@ TesselatedPolygon::drawTesselation(const PositionVector& shape) const {
     if (myTesselation.empty()) {
         myCurrentTesselated = this;
         // draw the tesselated shape
-        double* points = new double[shape.size() * 3];
+        size_t shapeSize = shape.size() * 3;
+        size_t holesSize = 0;
+        for (PositionVector hole : myHoles) {
+            holesSize += hole.size() * 3;
+        }
+        double* points = new double[shapeSize + holesSize];
```
behrisch, line 126: I would prefer if you use just one numPoints counter here and if you iterate over the holes with a const reference

bcoueraud87: > I would prefer if you use just one numPoints counter here and if you iterate over the holes with a const reference

I corrected that.

src/utils/gui/globjects/GUIPolygon.cpp
```diff
@@ -150,11 +153,22 @@ TesselatedPolygon::drawTesselation(const PositionVector& shape) const {
             gluTessVertex(tobj, points + 3 * i, points + 3 * i);
         }
         gluTessEndContour(tobj);
-
+        size_t startIndex = shape.size() * 3;
+        for (size_t j = 0; j < myHoles.size(); j++) {
+            PositionVector hole = myHoles[j];
```
behrisch, line 158: Sorry for not mentioning it earlier but the same remark of iterating with a const ref applies here too.

src/utils/gui/globjects/GUIPolygon.cpp
```diff
@@ -150,11 +153,22 @@ TesselatedPolygon::drawTesselation(const PositionVector& shape) const {
...
+            gluTessBeginContour(tobj);
+            for (int i = 0; i < (int)hole.size(); i++) {
+                points[startIndex + 3 * i] = hole[i].x();
+                points[startIndex + 3 * i + 1] = hole[i].y();
+                points[startIndex + 3 * i + 2] = 0.;
+                gluTessVertex(tobj, points + startIndex + 3 * i, points + startIndex + 3 * i);
```
behrisch, line 164: here as well. Please iterate over the vector and increment the startIndex in the loop

src/utils/shapes/SUMOPolygon.h
```diff
@@ -47,6 +47,9 @@ class SUMOPolygon : public Shape, public Parameterised {
     /// @brief friend class
     friend class PolygonDynamics;
 
+    /// @brief Default value for the vector holding the polygon's holes.
+    static const std::vector<PositionVector> DEFAULT_HOLES;
+
     /** @brief Constructor
```
behrisch, line 53: What is that one for?

src/utils/shapes/SUMOPolygon.h
```diff
@@ -109,17 +117,31 @@ class SUMOPolygon : public Shape, public Parameterised {
...
      * @param[in] geo  Whether to write the output in geo-coordinates
      */
     void writeXML(OutputDevice& out, bool geo = false) const;
 
+    /// @brief Return the exterior shape of the polygon.
+    PositionVector& getShapeRef();
+
+    /// @brief Return the holes shapes of the polygon.
+    std::vector<PositionVector>& getHolesRef();
+
```
behrisch, line 137: Why do you need to make the polygon internals modifiable?

src/utils/shapes/SUMOPolygon.h
bcoueraud87: I forgot to remove it I guess.

PR #13293, opened 2023-05-22
[TraCI] Added minPPET retrieval  

bcoueraud87:


m-kro:
Seems some more tests still have to be updated

PR #13327, opened 2023-05-27
[tools] tileGet - get tiles in parallel

dmfrodrigues:
Tile providers (namely ArcGIS) have high throughput but high latency. This means it takes a considerable amount of time to get one tile, but many tiles can be requested in parallel to increase throughput drastically.

I suggest using a thread pool to process multiple tile requests in parallel. The thread pool has 8 threads, which means at most 8 requests are being processed simultaneously.

Processing a tile is mostly server-intensive on the side of the tile provider, so in theory we could make all requests upfront in parallel and asynchronously wait for all the responses from the tile provider, which would still incur almost no CPU usage on the client side (since we're only getting relatively small images and storing them). But this would cause excessive load on the tile provider, and we could risk getting throttled by the tile provider. Hence why the thread pool only has 8 threads.

Still, this solution drastically reduces execution time for tileGet.py, which is especially beneficial when downloading large numbers of tiles to obtain a high-quality background for SUMO.

### DIFF
--- a/tools/tileGet.py
+++ b/tools/tileGet.py
@@ -81,6 +81,15 @@ def worker(options, request, filename):
     if os.stat(filename).st_size < options.min_file_size:
         raise ValueError("small file")
 
+def writeDecals(x, y, zoom, net, filename, decals, options):
+    lat, lon = fromTileToLatLon(x, y, zoom)
+    upperLeft = net.convertLonLat2XY(lon, lat)
+    lat, lon = fromTileToLatLon(x + 0.5, y + 0.5, zoom)
+    center = net.convertLonLat2XY(lon, lat)
+    print('    <decal file="%s" centerX="%s" centerY="%s" width="%s" height="%s" layer="%d"/>' %
+            (os.path.basename(filename), center[0], center[1],
+            2 * (center[0] - upperLeft[0]), 2 * (upperLeft[1] - center[1]), options.layer), file=decals)
+
 def retrieveMapServerTiles(options, west, south, east, north, decals, net):
     zoom = 20
     numTiles = options.tiles + 1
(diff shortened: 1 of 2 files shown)


behrisch:
Thanks for that one! I would rather have this optionally and introduce the number of threads as an option to the script. Can you change that?

tools/tileGet.py
```diff
@@ -82,17 +88,29 @@ def retrieveMapServerTiles(options, west, south, east, north, decals, net):
...
-                raise ValueError("small file")
+        
+    # opener = urllib.build_opener()
+    # opener.addheaders = [('User-agent', 'Mozilla/5.0')]
+    # urllib.install_opener(opener)
+    
+    with PoolExecutor(8) as executor:
+        futures = []
+
+        N = (ex - sx + 1) * (ey - sy + 1)
```
behrisch, line 99: this seems to be unused

dmfrodrigues: Ah yes. Because I was using it to track progress in line 112
```py
print(f"{(100.0 * (i+1)/N):.2f}% ({request})")
```
but I ended up commenting it out. I'll comment this one out as well, but won't remove it since it might be useful at some point (and also if you're running it manually, you may uncomment them to check download progress)

dmfrodrigues:
> Thanks for that one! I would rather have this optionally and introduce the number of threads as an option to the script. Can you change that?

@behrisch how about a CLI option with number of threads, which defaults to 0 (no parallelism), and if >0 a thread pool with that size is created?

behrisch:
That would be great. I would also not mind to have a default of 8. And can we maybe make it backward compatible with python2?

dmfrodrigues:
> That would be great. I would also not mind to have a default of 8. And can we maybe make it backward compatible with python2?

- Implemented number of threads as CLI option, default is 8.
- Backward-compatible with python2 using `multiprocessing.pool.Pool`

Also commented out unused code.

@behrisch could you take another quick look?

tools/tileGet.py
```diff
@@ -82,25 +89,48 @@ def retrieveMapServerTiles(options, west, south, east, north, decals, net):
...
+    # opener.addheaders = [('User-agent', 'Mozilla/5.0')]
+    # urllib.install_opener(opener)
+    
+    # N = (ex - sx + 1) * (ey - sy + 1)
+    
+    if options.parallel_jobs != 0:
+        original_sigint_handler = signal.signal(signal.SIGINT, signal.SIG_IGN)
+        pool = Pool(options.parallel_jobs)
+        signal.signal(signal.SIGINT, original_sigint_handler)
+
```
behrisch, line 103: What is the signal handling for?

tools/tileGet.py
```diff
@@ -202,11 +234,11 @@ def get(args=None):
...
-        if e.code == 403:
-            print(" Maybe an API key is required.", file=sys.stderr)
-    except ValueError as e:
-        print("Error: Tile server returned %s." % e, file=sys.stderr)
+    # except urlerror as e:
+    #     print("Error: Tile server returned %s." % e, file=sys.stderr)
+    #     if e.code == 403:
+    #         print(" Maybe an API key is required.", file=sys.stderr)
+    # except ValueError as e:
+    #     print("Error: Tile server returned %s." % e, file=sys.stderr)
```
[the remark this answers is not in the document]
dmfrodrigues: My mistake.

tools/tileGet.py
behrisch: Just out of curiosity. This does not happen with the previous implementation with concurrent.futures?

tools/tileGet.py
```diff
@@ -82,25 +89,48 @@ def retrieveMapServerTiles(options, west, south, east, north, decals, net):
...
+        # print('%f%% (%s)' % (100.0 * (i+1)/N, request))
+        if net is not None:
+        
+            lat, lon = fromTileToLatLon(x, y, zoom)
+            upperLeft = net.convertLonLat2XY(lon, lat)
+            lat, lon = fromTileToLatLon(x + 0.5, y + 0.5, zoom)
+            center = net.convertLonLat2XY(lon, lat)
+            print('    <decal file="%s" centerX="%s" centerY="%s" width="%s" height="%s" layer="%d"/>' %
+                    (os.path.basename(filename), center[0], center[1],
+                    2 * (center[0] - upperLeft[0]), 2 * (upperLeft[1] - center[1]), options.layer), file=decals)
```
behrisch, line 134: Does this really work for the non-parallel case? It seems to me futures will be empty then so no processing will take place

tools/tileGet.py
dmfrodrigues: No it did not happen. `concurrent.futures` works just fine without the hack

tools/tileGet.py
dmfrodrigues: It does. Futures are only inserted if the pool has threads. If the pool is empty, the worker code is run in line 113 without creating futures.

Tho I've just noticed a small bug in line 113 related with unnecessary arguments, will fix.

behrisch: Yes, the images will be retrieved but the decal lines will not be printed, will they?

behrisch:
Your solution duplicates quite some code. Could you either refactor it into a function or simply add a dummy `None` future in the non-parallel case?

behrisch:
Thanks a lot!

PR #13473, opened 2023-06-21
Fix some incorrect TraCI documentation

OtterBas:


namdre:
Thanks for the careful look!

Are you building your own TraCI client? If so, what language are you targeting?

OtterBas:
I'm working with Eclipse MOSAIC, so I'm using Java

namdre:
are you aware of libsumo/libtraci for java?
https://sumo.dlr.de/docs/Libtraci.html#java
https://sumo.dlr.de/docs/Libsumo.html#java

OtterBas:
I was not aware of that, so thanks for pointing it out.
However, since I will also need to include V2X messages in the simulation I'm using MOSAIC.
This handles running the SUMO simulation and I'm only using TraCI to dynamically change the behavior of individual vehicles.
There is already built-in support for accessing the TraCI, however it requires building the commands manually, which is where I ran into these erros in the documentatation

PR #13625, opened 2023-07-14
[JuPedSim] Improved FCD output.

bcoueraud87:


src/microsim/transportables/MSPModel_JuPedSim.h
```diff
@@ -80,26 +80,33 @@ class MSPModel_JuPedSim : public MSPModel {
         Position getPosition(const MSStageMoving& stage, SUMOTime now) const;
         void setPosition(double x, double y);
 
+        Position getPreviousPosition() const;
```
behrisch, line 83: This function gets never called

src/microsim/transportables/MSPModel_JuPedSim.h
```diff
@@ -114,7 +121,10 @@ class MSPModel_JuPedSim : public MSPModel {
         Position myDestination;
         JPS_AgentId myAgentId;
         Position myPosition;
+        Position myPreviousPosition; // Will be initialized to zero automatically.
         double myAngle;
+        double myLanePosition;
+        MSEdge* myCurrentEdge;
```
behrisch, line 127: This seems to be unused

src/microsim/transportables/MSPModel_JuPedSim.cpp
```diff
@@ -696,12 +723,13 @@ SUMOTime MSPModel_JuPedSim::PState::getWaitingTime(const MSStageMoving& /* stage
 
 
 double MSPModel_JuPedSim::PState::getSpeed(const MSStageMoving& /* stage */) const {
-    return 0;
+    Position velocity = (myPosition - myPreviousPosition) / STEPS2TIME(DELTA_T);
+    return velocity.length2D();
 }
```
behrisch, line 728: This is quite inefficient. You could just use myPosition.distanceTo2D(myPreviousPosition) / TS

behrisch (changes requested):

src/microsim/transportables/MSPModel_JuPedSim.h
bcoueraud87: Yes. I removed it, there is no need to expose this attribute.

src/microsim/transportables/MSPModel_JuPedSim.h
bcoueraud87: `myCurrentEdge` has been removed, thanks.

PR #13740, opened 2023-08-24
Fixes regression in #12990

mschrader15:
The implementation of red transfer in #12990 led to edge cases where illegal barrier cross could occur. The problem was incorrect transition time calculation for lights in red transfer state.

I simplified the red transfer logic and corrected the error in transition time calculation for phases in red transfer (`0`)

PR also includes refactoring & ran `astyle`.


behrisch:
Sorry for not looking at this one earlier. There was a lot of empty line removal which is something astyle does not (at least on our side) Could you please revert this if possible?

mschrader15:
Hey @behrisch I'm struggling to implement the whitespace revert via git-diff. Beyond my skill level. 

I can manually diff if the whitespace/line removal issue is critical.

mschrader15:
I removed whitespace / newline changes to the best of my ability

namdre:
Thanks for the reminder ;-)

Issue #13681, opened 2023-08-02
Distance to leader when using ACC car-following model 


adoreador:
Hello, I have some questions about the edgedata in my simulation. I have set up a three-lane simulation with ACC (Adaptive Cruise Control) as the car-following model. The vehicle length is 4.75 meters, and I have set tau="1.0", mingap="1.5", and desiredMaxSpeed="15.39". This configuration resulted in a stable traffic flow, and I have set the edgedata frequency to 10 seconds. 
In the output file, I obtained the values density="98.18", laneDensity="32.73", and speed="15.27". However, when I calculated the laneDensity as 1000 / Headway, using Headway = length + minGap + tau * speed, I noticed that it does not match the previously obtained value. I'm a bit puzzled by this discrepancy and would greatly appreciate any insight you could provide.

Thank you for your assistance.

robertalms:
Thanks for your question, but densities is a complex matter, so I fear your question is not precise enough to explain your discrepant expectation adequately.

1. It would be very helpful, if you provided an example to replicate the scenario with your edgedata outputs.
2. Your rough estimation of a lane density is higher than the edgeData laneDensity, if I understand you correctly?! Mind, that you'd have to configure multiple things in your scenario to validate your expected outcome: [see doc](https://sumo.dlr.de/docs/FAQ.html#how_do_i_insert_vehicles_with_a_fixed_density). 
3.  So far, I don't see that your issue is specifically related to the ACC model, but be aware that the ACC should run with lower step-length 0.1 (which effects the lane density estimation.) [see doc](https://sumo.dlr.de/docs/Car-Following-Models/ACC.html#notes)

Maybe also have a look at [doc](https://sumo.dlr.de/docs/Simulation/RoadCapacity.html#further_headway_effects) for further insights about the impacts of the vehicle parametrization on the vehicle insertion and consequent traffic density.

adoreador:
Thank you for your response. I will proceed to read the document for further learning. By the way, may I continue to ask questions here if I have any in the future?

robertalms:
Sure, feel free to ask further questions related to the topic here.

@m-kro will close the issue for now, but we can re-open it anytime, in case you/we discover any problems with the lane density related to the ACC model.

adoreador:
Hello, I have studied the documentation and reviewed my code, but I still have some questions. Below is a portion of my code from the rou file.
```
<routes>
[...]

PR #13755, opened 2023-08-29
Fix for ACC model - eliminate vehicle length from spacing margin

robertalms:
see #13681


behrisch:
I suppose this is referring to equation 5 in https://pure.tudelft.nl/ws/portalfiles/portal/51734038/1_s2.0_S0968090X1830528X_main.pdf. You are right the L is superfluous and should not be in the code. But I have two questions: Why did you remove the "- 5". And do you have any idea what to do about the minGap here? I did not read the paper completely. Do they assume that such a concept exists or is it just part of the vehicle length?

robertalms:
1. Sorry, removing the ' - 5' from the equation was a mistake on my side. Thanks for double checking!
2. No, I don't think they account for a safe minimum distance as `minGap`. Neither in the original publication nor in this one is anything mentioned about something like this. So, I think it's indeed bumper to bumper distances. I guess, this is why they added this spacing margin later on.

behrisch:
But the original reporter of #13681 tried to account for the minGap so I thought it would be fair to do so as well. gap2pred in the current code does not include the minGap so it is not bumper to bumper. But you are probably right If we allow the model to violate the minGap this is a sure way to more collisions because sumo already reports minGap violations as such

robertalms:
please also reopen #4551

PR #13793, opened 2023-09-06
SumoAnka the ui modifications

mfsen:


behrisch:
Thank you! Is the whole mono stuff really necessary?

behrisch:
@palvarezlopez Can you please check whether it would be possible to integrate the netedit changes as a new view?

mottmelissa92-ops (approved):

behrisch:
@mfsen I don't think we will merge this but please get back to us if there are still parts of the UI you would like to see being changed in the mainline

PR #13801, opened 2023-09-07
Corrected the docs and disabled GEOS message handler

bcoueraud87:


behrisch:
I think the BUILD_JPSVIS variable is not used anymore so we should remove it

behrisch:
Please do not mix docs and unrelated code changes. Use a different PR for that. Also if you remove the message handler, you can also remove the message handler function

PR #13814, opened 2023-09-12
[JuPedSim] Took into account projection param

bcoueraud87:


palvarezlopez:
part of https://github.com/eclipse/sumo/pull/13809

PR #13832, opened 2023-09-18
Path to Userdoc not found

Enough7:
Hi,
I could not found said path but I think the path given by the commit is the correct path

### DIFF
--- a/docs/web/docs/index.md
+++ b/docs/web/docs/index.md
@@ -14,7 +14,7 @@ If You use SUMO, please tell us about your
 The content of this Documentation is freely editable. In order to edit a certain page, click on the "Edit on GitHub" button on the upper right corner and submit a Pull Request.
 Here is a short help on [editing articles](Editing_Articles.md). You can also [Build](Developer/Documentation_Build.md) the documentation locally, or [Download](https://sumo.dlr.de/sumo_documentation.zip) a copy.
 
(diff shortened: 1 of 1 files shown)


m-kro:
Actually the current path should be correct, as you can check easily by downloading a release from our [downloads section](https://sumo.dlr.de/docs/Downloads.php), e.g. [v.1.18.0 for Windows](https://sumo.dlr.de/releases/1.18.0/sumo-win64-1.18.0.zip). Once unzipped, it contains subfolder docs/userdoc/ where all HTML doc pages as they were at release time are available.

Enough7:
Why does a local build not include those docs?
I think it would be nice if every build would be on the same page.
The same goes for doxygen,
I only could find the 'daily-version' of the doxygen docs [here](https://sumo.dlr.de/daily/doxygen/) but I am interested in my 'local-version' and perhaps in older versions of the doxygen docs.
Where can I find them?

m-kro:
You can generate both on your system, see [documentation build](https://sumo.dlr.de/docs/Developer/Documentation_Build.html) and [Doxygen website](https://www.doxygen.nl/) (the doxygen config sumo.doxyconf is included on root level of the SUMO repository).

Enough7:
@m-kro Thank you for clarification

PR #13834, opened 2023-09-18
[JuPedSim] Improved options for drawing JPS pedestrian network

bcoueraud87:


behrisch:
I think reading the new settings is still missing?

PR #13908, opened 2023-10-05
HYL ：Custom lane change model-AL2023

huyanlei123:
I have customized the lane changing model according to the official website, but there may be errors“Error: Unknown lane change model 'AL2023' when parsing vType 'CarA'Quitting (on error).”


m-kro:
Please do not issue a pull request just for asking questions. You can still keep the code in your fork and add a link to it in the issue.

PR #13975, opened 2023-10-23
Dockerfile for Alpine Linux

vijay-jaisankar:
## Main features of this PR

- Dockerfile using Alpine as the base image
- Finding replacements and workarounds to install deps to build SUMO
- Instructions to build this image

### Notes
- To add `fox-dev` to the `apk` environment, edge installations need to be enabled as `fox-dev` is [not part of the main packages list](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/fox-dev).
- `bash` and `make` need to be preloaded in separate steps to maintain usage compatibility with other images; and [to enable CMake builds](https://stackoverflow.com/questions/55190714/unable-to-use-make-in-alpine-image); respectively.
- A python virtual environment is required to resolve environment clashes with `apk`.

A test build(`Docker version 20.10.23, build 7155243`) was successful: [Dockerhub](https://hub.docker.com/r/thehungrypigeon/alpinesumo)


behrisch:
Thanks for this one! Can you share where you use it?

PR #13995, opened 2023-10-26
Implemented getIntervalVehicleNumber in InductionLoopScope for the Traci-API on the client-side

Enough7:
Hey,

I needed this for Artery and I would like to share this with you.
I am not familiar with your test-setup so I did only a little testing by hand.

### DIFF
--- a/src/libsumo/TraCIConstants.h
+++ b/src/libsumo/TraCIConstants.h
@@ -704,9 +704,6 @@ TRACI_CONST int AUTOMATIC_CONTEXT_SUBSCRIPTION = 0x03;
 // generic attributes (get/set: all)
 TRACI_CONST int GENERIC_ATTRIBUTE = 0x03;
 
-// The number of vehicles (or persons, if so configured) that passed the detector during the current interval
-TRACI_CONST int INTERVAL_NUMBER = 0x25;
-
 // last step vehicle number (get: induction loops, multi-entry/multi-exit detector, lanes, edges)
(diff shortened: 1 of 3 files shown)


behrisch:
Thanks for the PR! The old TraCIAPI is officially deprecated so I am a little hesitant to modify it at all. Is there any reason for not using libtraci? Also why did you add another constant to TraCIConstants.h? There is already VAR_LAST_INTERVAL_NUMBER.

Enough7:
Hi @behrisch,
thanks for the feedback.

> The old TraCIAPI is officially deprecated 

I did not know that, is there a new Traci-API or is it libtraci?

> Is there any reason for not using libtraci?

I guess [Artery](http://artery.v2x-research.eu/) does not use it?

> Also why did you add another constant to TraCIConstants.h? There is already VAR_LAST_INTERVAL_NUMBER.

I did not notice.

behrisch:
@Enough7 Could you still do the "paperwork" (make an eclipse account and agree to the ECA), see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work?

Enough7:
@behrisch I use an E-Mail-Mask given by Github, which I can not use to receive E-Mails afaik. So I guess the CICD-Script would not recognize my commit's E-Mail-Address anyway. I can assure you that me and my commits apply to the ECA found [here](https://www.eclipse.org/legal/ECA.php). I may sign something with my SSH-Key I use for GitHub.

PR #14047, opened 2023-11-07
compute penalty for rejecting requests

ritzerp:
- previously the penalty had a default value of 10000
- now the penalty depends on the dimension of the network (diameter of bounding box)  and a penalty_factor
- penalty = penalty_factor * dimension_of_network
- the penalty factor can be modified by user input
- if cost type is time, a default speed is assumed to get equivalent dimensions


behrisch:
@ritzerp Thank you! Can you add a test? Maybe in a separate PR?

behrisch:
@rummel123 Can you have a look?

rummel123:
> @rummel123 Can you have a look?

Yes

ritzerp:
> @ritzerp Thank you! Can you add a test? Maybe in a separate PR?

Yes, good idea. I will try to add a test in a separate PR in the next weeks.

rummel123:
Looks good, can be merged.

behrisch:
@ritzerp There are several purely cosmetic changes. Could you revert them so that we can focus on the content?

ritzerp:
@behrisch I reverted the cosmetic changes. Sorry for the late reply!

ritzerp:
I think I found a small bug with my PR and will solve it... Please wait for the push.

ritzerp:
@behrisch solved the problem now. The penalty has to be an integer.

behrisch:
Thanks a lot! I still see some places where whitespace gets added at the end of a line. Can you try to remove them?

behrisch:
Thanks a lot!

Issue #14139, opened 2023-12-04
Jitter the position of passengers that get out of a train.


bcoueraud87:
Passengers of a train go out of the carriage all aligned. Jitter this default position a little bit.

bcoueraud87:
@behrisch Please assign me.

PR #14140, opened 2023-12-04
Jittered position of outgoing train passengers (#14139).

bcoueraud87:
I used a uniform random variable but can do something else if really needed. Do we have access to a default carriage door width? Do we make this somehow configurable (does everybody wants this)?


behrisch:
I think uniform is fine for a start but we check the result against the lane length

PR #14303, opened 2024-02-03
Update macrOutput file

Amirhoseinkr73:
the owner of the file is me, and I want to commit the change to have it on my account also.


m-kro:
The script has been updated a few times since the initial commit for several reasons:

- use argparse
- compatibility with shared plotting settings of the other visualisation scripts
- output directory
- typos
- etc...

These changes would be removed by your commits. A pull request is meant for adding some value, not the opposite. There are two things which we still may adopt if you provide a clean (new) pull request (meaning no merge or revert commits):

- spelling of your name (you changed it in one commit but then overwrote it in another?!)
- some comment on how to use the script / literature references (please cross-check with the current script version)

Amirhoseinkr73:
Hi again, I changed the date to its creation date as you mentioned.
Best regards,
Amir

PR #14319, opened 2024-02-13
[JuPedSim] Refactored code related to JPS_GeometryBuilder

bcoueraud87:


behrisch:
This one removed all my code about the parameters. I will not revert it since I wanted to change that anyway but please @bcoueraud87 have a second look whether it undid some other things as well!

Issue #14334, opened 2024-02-15
Unboard train's passengers through doors when the rail line is curved


bcoueraud87:
The current code, located in MSStageDriving.cpp, makes the assumption that the train is driving on straight rail line. Thanks to the train helper, it is possible to now take into curvature the curvature.

bcoueraud87:
Please assign me.

bcoueraud87:
Note: the original feature was created for the needs of the JuPedSim coupling but actually can be used standalone.

namdre:
Would you mind listing the pull requests or direct commits that already went into this ticket?

bcoueraud87:
I think we should close this ticket, even if #14335 and #7049 are open.

PR #14336, opened 2024-02-15
Unboard train's passengers through doors when the rail line is curved (#14334)

bcoueraud87:


namdre:
can you explain why an existing test changed it's behavior? The rail line in that test is not curved and also the door attribute isn't used in the scenario.

namdre:
Sorry, I missed f9b6e6855962d even though you linked it in #14336 and mistook the test for another one I knew.

Issue #3920, opened 2018-03-19
Provide possibility for non-constant acceleration profile in standard model


leoluecken:
The main reason for the introduction is to accurately model slow starting after a vehicle resumes from a halt.
I'd propose to enable switching to Gipps' acceleration profile (for the maximal acceleration)
a(v(t)) = 2.5*a_max*(1-v(t)/v_max)*sqrt(0.025 + v(t)/v_max),
where v_max is the desired travelling speed according to Gipps (this will induce discontinuous acceleration profiles where the speed limit changes, but this is no big issue to my opinion). However, one can argue that the value for v_max should rather be the maximal physical speed of the vType. The coefficient a_max equals the vType's accel parameter.
The switch may be binary (Gipps profile on/off) or continuous with parameter h in [0,1]:
a_model(v) = h*a(v) + (1-h)*a_max

leoluecken:
We decided to use v_max as maximal speed as defined by vType, since using the lane specific allowed speed would imply that the velocity profile at starting would strongly depend on the allowed speed on the lane, which seems implausible.

leoluecken:
Peter proposes that the user should be able to provide a lookup table for a custom profile in some way.

namdre:
A lookup table would be useful. Something like 
accelProfile="0,3 5,2.5 20,2 30,1" to model a piecewise linear profile with acceleration of 3 at speed 0 and acceleration 1 at speed 30 (similar for decelProfile).

This could be used to model different things:
a) available power for acceleration in different gears (maximum possible acceleration)
b) slow-to-start behavior (desired acceleration)
The interaction with slope would be different: grade directly affects a) but not b)

This might make it necessary to distinguish `desiredAccelProfile` and `maxAccelProfile`

Domsall:
We would like to use this feature for the EIDM.
I would implement it, but need some advice regarding the interaction with `amax`.

When we add `maxAccelProfile` to the list of parameters, should `amax` be ignored? Or should `amax` be used as a parameter and a profile, depending on the users input?

namdre:
I'd prefer having a new attribute and then use `accel` as an upper bound on acceleration (on top of the defined curve)

PR #14339, opened 2024-02-15
Add maxAccelProfile and desAccelProfile to CF-models

Domsall:
Seperated `maxAccel` (physically possible acceleration) from `desAccel` (speed-dependent, non-constant myAccel).

Both profiles need to be written in the form of `"speed1 accel1, speed2 accel2, speed3 accel3, ..."`.

It works/is intended to work as follows:

- (not implemented in any model yet) `maxAccelProfile`: These acceleration values represent the physically possible acceleration at a specific speed. If a CF-model calculates the possible acceleration depending on vehicle resistance, then these values shall be used. When the maxAccel-value becomes lower than the desAccel value, the actual accel-value gets reduced.
- (used in the EIDM, can easily be adapted for any other model) `desAccelProfile`: These acceleration values represent the desired acceleration at a specific speed. As "dummy" value, the acceleration output is equal to `myAccel`.

Solves #3920 for the EIDM, but needs to be added separately for the other models.


Domsall:
Same here: Can you estimate when this will be integrated?

namdre:
> Same here: Can you estimate when this will be integrated?

This won't make it into 1.21.0 sorry. 1.22.0 should be doable.

Domsall:
no problem, thanks for the update! I published a paper using this feature and just wanted it to be available for everybody.

namdre:
@Domsall I realize this is a bit rushed but I would like to integrate this rather than wait another release.

In the meantime we introduced a similar concept of speed-based tables for railway simulation (https://sumo.dlr.de/docs/Simulation/Railways.html#custom_dynamics_model_with_tabular_data)
Do you mind terribly if I rewrite the data ingestion so that it loads a `speedTable` and change the semantics of `maxAccelProfile` and `desAccelProfile` so it only holds a list of accelerations rather than a list of pairs?

Domsall:
yes, you can absolutely change the semantics from a list of pairs to two seperate variables/vectors

namdre:
@Domsall by looking at the code, I assumed that the behavior of our EIDM tests should remain unchanged since the default profiles return `myAccel` at all speeds. 

However, in line `MSCFModel_EIDM::patchSpeedBeforeLCEIDM` you replaced the `MIN3(.., 1.5 ...)` with a `MIN2(...)` which changes the `intensity` calculation with the default constant desAccelProfile.

was this intended?

Domsall:
The MIN3-commit actually came later ;-) MIN2 is the older version, so please use the MIN3

PR #14373, opened 2024-02-20
getMaxTLLinkIndex in a node

aminissn:
This is one of the many required elements to implement empirical traffic signal logs in SUMO


namdre:
Thanks! I'm looking forward to your signal log code as well.

PR #14379, opened 2024-02-21
[JuPedSim] Corrected JPS_VERSION definition

bcoueraud87:


behrisch:
Thanks for the idea, I fixed it in https://github.com/eclipse-sumo/sumo/commit/58f1b48dcc1e7eca20c62e9874ae1b4142272171 because the windows build was still failing. It will also break if Jupedsim ever decides to go up to a patchlevel of 10 (that's why the 0 has been introduced). But let's hope it won't happen.

PR #14395, opened 2024-02-23
Plexe: Improved whole-platoon lane change procedure

michele-segata:
This pull request improves the lane change behavior of a whole platoon, which was known to cause collisions in some specific cases. In the updated code, the leader checks for lane change safety conditions by invoking the `getNeighbors()` method with the safety bit enabled. Only if all members agree on lane change safety the maneuver is performed.

In addition, a new API to safely move the platoon to a specific lane upon user request has been implemented.


namdre:
Happy to see you're still working on this!

michele-segata:
Oh yes, I'm always trying to do so! :) Unfortunately not much time is available, but I'm trying to keep stuff up to date!

PR #14499, opened 2024-03-13
More vehicle XML files for the MMPEVEM.

kevbad:
Hello everyone!

I've created some more vehicle XML files for the [MMPEVEM](https://sumo.dlr.de/docs/Models/MMPEVEM.html).

Yours,
Kevin


behrisch:
Also thanks from my side! If you plan to add more files please do not set action step length and color. It may have undesired side effects for the user. Looking forward to a Tesla Model Y file if you can create any :-)

PR #14564, opened 2024-03-20
[JuPedSim] Added grid for passengers unboarding and constraint resolution

bcoueraud87:


namdre:
SUMO.ci test-case basic/personFlow/unboarding_curved fails

PR #14662, opened 2024-04-04
[JuPedSim] Changed shuffle code to be platform independent

bcoueraud87:


bcoueraud87:
@behrisch: The difference between the + operator and the [ ] operator is that the second automatically dereferences the iterator.

PR #14749, opened 2024-04-19
[JuPedSim] Changed geometry generation to remove bulges

bcoueraud87:
- Networks in tests have been upgraded with netconvert to take into account the new outline shapes for crossings.
- Dusseldorf scenario works.


behrisch:
the larger_net test still fails

Issue #13024, opened 2023-04-12
Lane occupancy calculation bug


tasosb:
Hello,

This error occurs for micro-simulation and for vehicles that have their vehicle type changed, while they are moving on a lane. More specifically, in  `microsim/MSLane.cpp`, the brutto occupancy depends on the value `myBruttoVehicleLengthSum` (veh_length+min_gap). This value is only updated upon entry and exit of a vehicle in the lane, disregarding the fact that the vehicle's length and minGap value can be changed as the vehicle is in the lane, either directly by traci or by changing it's vType. For example, the vehicle's length+min_gap value at the lane exit may be lower than these values at the lane entry, leaving the difference as residue to the value `myBruttoVehicleLengthSum`. After some time, this leads to an inaccurate 100% occupancy of the lane and no more vehicles are able to join the lane.

![Untitled](https://user-images.githubusercontent.com/18355854/231588736-0abafd42-7b10-472f-bd8c-71f1db559ca7.png)

A possible solution could be to calculate `myBruttoVehicleLengthSum` and `myNettoVehicleLengthSum`in `getBruttoOccupancy` and `getNettoOccupancy` directly, by iterating the list of vehicles in the lane.

**SUMO-version:**
v1.16 and git build

**operating system:**
Windows 10 & Ubuntu

namdre:
iterating the list of vehicles is probably to inefficient. 
Since changing the type of a vehicle is presumably a rare event, updating the occupancy upon type change sounds like a better solution. 

Out of curiosity, are you actually changing vehicle lengths or just minGaps at runtime? If the former, what are you modelling?

tasosb:
Actually the issue came up when using simpla, that generally does a lot of vType changing. More specifically I set a relatively lower min_gap for the follower, catchup and/or catchup follower vTypes than the leader and default vType. In almost 1hour of simulation, the accumulated mingap difference was enough to flag the lane as having reached it's maximum occupancy.

I am only changing the minGaps, not the length, I just mentioned it as there is a relative setLength function that someone could use for a reason.

PR #14820, opened 2024-05-04
Fix for wrong brutto occupancy calculation when minGap of vehicles is changed 

tasosb:
This PR solves the problem with variable minGaps (e.g. for platoons) causing a wrong brutto occupancy calculation (see #13024) and affecting vehicle insertion.

minGap can be changed in the following ways (possibly not an exhaustive list):

1. Changing minGap for a single vehicle, which creates a singular VType
2. Changing minGap for a VType
3. Changing the VType for a vehicle (happens in simpla)

Case 1 and Case 3 can lead to the problem, especially if used in a loop for multiple vehicles, as the exiting minGap is different than the entering minGap
Case 2 is less serious as the minGap is changed to the same value for both the inserting and exiting vehicles and any accumulated minGap difference (vehicles already moving when the change is happening) is static

The proposed solution solves the problem for the severe cases 1 and 3 by setting a flag to the vehicle's current lane object that triggers a brutto occupancy recalculation in the next step. 

Despite the existence of a setMinGap function in the libsumo Vehicle C++ API, it was not set in the TraCI Server API and was redirected to the VehicleType C++ API setMinGap function  (https://github.com/eclipse-sumo/sumo/blob/main/src/traci-server/TraCIServerAPI_Vehicle.cpp#L1404). With the proposed solution, the already existing Vehicle setMinGap function is added to the TraCI Server API, so that the MSVehicle object can access the lane object and set the recalculation flag.

Before fix:
image

After fix:
image

### DIFF
--- a/src/libsumo/Vehicle.cpp
+++ b/src/libsumo/Vehicle.cpp
@@ -1927,6 +1927,7 @@ Vehicle::setType(const std::string& vehID, const std::string& typeID) {
     MSVehicle* microVeh = dynamic_cast<MSVehicle*>(veh);
     if (microVeh != nullptr && microVeh->isOnRoad()) {
         microVeh->updateBestLanes(true);
+        microVeh->updateLaneBruttoSum();
     }
 }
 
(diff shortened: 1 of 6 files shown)


namdre:
Thanks a lot! 
I'd love to merge this directly but I'd need you to click through the Eclipse Contributor Agreement first. See https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work

PR #15066, opened 2024-06-20
Add upper boundary to drivererror-intensity in EIDM

Domsall:
Vehicles with high `myAccel` might calculate high `drivererror` values.
This is unintended. An upper boundary solves this issue.

The intensity value is mainly needed for vehicles with low `myAccel`.


namdre:
maybe this upper boundary should itself be configurable?

Domsall:
One can still create higher drivererrors with a higher sigmaerror. From my point of view, this should just not depend on `myAccel`. 

The error should only depend on `myAccel`, if `myAccel` is low. If not, the error could become higher than the intended acceleration itself.

PR #15067, opened 2024-06-20
Change the GLOSA adaptSpeed calculation

Domsall:
I changed 3 main parts of the adaptSpeed calculation:

1. As pointed out in https://github.com/eclipse-sumo/sumo/issues/12217, targetSpeed can be higher than currentSpeed. This is considered now.
2. As pointed out in https://github.com/eclipse-sumo/sumo/issues/15061, targetSpeed was calculated similar to the green curve. Now it follows the orange curve.
3. Instead of using `setSpeedTimeLine`, I added `setChosenSpeedFactor`.


namdre:
Thanks!

Issue #15048, opened 2024-06-18
sumo to xodr question: signal not on road


lindsaymorgan:
I used netconvert to convert .net.xml to xodr, then used Roadrunner to visualize xodr file. I found most of those signals are not on road. Is there any method to solve this?
![image](https://github.com/eclipse-sumo/sumo/assets/25105493/b8e55d2a-cca3-4d55-8871-6484c208f656)

namdre:
can you provide a minimal input example of a .net.xml  and then explain why the signal is wrong? The signals written to .xodr are meant to represent the "colored lights" part so they are raised 5m of the ground. Or do you mean something else?

lindsaymorgan:
Hello~Sorry for the late reply. I just figure out why this happens. I built a simplest network here.
[example.zip](https://github.com/user-attachments/files/15924230/example.zip)
![image](https://github.com/eclipse-sumo/sumo/assets/25105493/bd0ec29a-6347-4bbb-b83c-ec0d9a3e5faf)
If I use netconvert to convert it into xodr and import it to Roadrunner, the signal will station far away from the actual road.
![image](https://github.com/eclipse-sumo/sumo/assets/25105493/36fbeb06-99d3-451a-86a2-152b1faaaf41)
And the Roadrunner output is as follow:
![image](https://github.com/eclipse-sumo/sumo/assets/25105493/98c8c781-dd54-4dbb-9056-7557c7b54960)
I read netconvert code and realize it is caused by precision difference between link length (8) and signal s,t (2). And also, there is a problem about the lane validity in the output xodr file. 
I have figured out how to fix this bug and will send a pr in 1 or 2 days.

PR #15078, opened 2024-06-22
fix xodr signal precision and lane validity; fix xodr connection with 0 length

lindsaymorgan:
Hello~ Here is my solution to solve issue #15048 .


namdre:
Thanks!

PR #15152, opened 2024-07-04
Additional GLOSA options

Domsall:
I added three additional options for GLOSA and made some changes to EIDM/MSVehicle-Code.

1. `device.glosa.add-switchtime`: When the vehicle reaches the intersection just before the light signal turns green, it may brake due to safety and the still active red light. With this added time, the vehicle reaches the intersection X seconds later and will not have to brake (but is farther away from the intersection when the light turns green).
2. `device.glosa.override-safety`: When this is true, the light signal is ignored (stopSpeed is calculated with a distance of infinity). The vehicle will always follow GLOSA's predicted speed (turns off the braking behavior from above).
3. `device.glosa.ignore-cfmodel`: When this is true, the "triangle" speed behavior is adapted via the previously used speedTimeLine (perfect speed calculation without errors).

### DIFF
--- a/src/microsim/MSVehicle.cpp
+++ b/src/microsim/MSVehicle.cpp
@@ -59,6 +59,7 @@
 #include <microsim/devices/MSDevice_Taxi.h>
 #include <microsim/devices/MSDevice_Vehroutes.h>
 #include <microsim/devices/MSDevice_ElecHybrid.h>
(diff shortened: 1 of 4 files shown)


namdre:
it would be great if you could also supply some tests for the new behavior

PR #15179, opened 2024-07-09
fix libsumo tls

aminissn:


namdre:
I think Micha fixed it already in 4e1b292d3a167eb15df1cb459e85335ec68c96fb

Or is it still failing for you after this?

behrisch:
I am not sure how important this subparameter thing really is but it might be broken after this

namdre:
I think it was broken before because the TraCIDefs constructor doesn't accept subParameters.

PR #15227, opened 2024-07-17
sumolib updates 

aminissn:
Some updates for sumolib that help to define bus priority rules in tllogic

### DIFF
--- a/data/po/de_sumo.po
+++ b/data/po/de_sumo.po
@@ -12915,7 +12915,7 @@ msgstr "falsch zugeordnete Zeile der Matrix A beim Lösen der Schaltung"
 
 #: src/utils/traction_wire/Circuit.cpp:351
 #: src/utils/traction_wire/Circuit.cpp:476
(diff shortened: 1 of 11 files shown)


namdre:
- please make two distinct pull requests for the sumolib changes and the libsumo changes
- please check whether the added E1 code is even needed in light of sumolib/sensors/inductive_loop.py

aminissn:
> * please make two distinct pull requests for the sumolib changes and the libsumo changes
> * please check whether the added E1 code is even needed in light of sumolib/sensors/inductive_loop.py

one quick question: I guess InductiveLoop, e1Detector and E1Detector are all the same. Does it make sense to make the naming unique?

namdre:
> one quick question: I guess InductiveLoop, e1Detector and E1Detector are all the same. Does it make sense to make the naming unique?

Yes, they are all the same. The preferred name for new code is `inductionLoop`.

PR #15230, opened 2024-07-17
update sumolib. with Condition and earlyTarget in Phase

aminissn:
I added my python code under tools/tls with all necessary changes in sumolib init

### DIFF
--- a/tools/sumolib/net/__init__.py
+++ b/tools/sumolib/net/__init__.py
@@ -162,9 +162,13 @@ class TLSProgram:
         self._offset = offset
         self._phases = []
         self._params = {}
+        self._conditions = {}
 
-    def addPhase(self, state, duration, minDur=-1, maxDur=-1, next=None, name=""):
-        self._phases.append(Phase(duration, state, minDur, maxDur, next, name))
+    def addPhase(self, state, duration, minDur=-1, maxDur=-1, next=None, name="",earlyTarget=""):
+        self._phases.append(Phase(duration, state, minDur, maxDur, next, name, earlyTarget))
+
+    def addCondition(self,id,value):
+        self._conditions[id] = value
 
     def toXML(self, tlsID):
         ret = '  <tlLogic id="%s" type="%s" programID="%s" offset="%s">\n' % (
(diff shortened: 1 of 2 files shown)


namdre:
- Please put busPriority.py into a separate pull request (it's not quite ready for inclusion, while the rest of the code is fine)
- Please add the new vTypes argument in induction_loop.py at the end of the argument list to ensure backward compatibility with existing code (default argument should be the empty string).

aminissn:
hopefully this is fine, even though the failing eca test bother me every time!

autumnfound:
> hopefully this is fine, even though the failing eca test bother me every time!

@aminissn I missed it due to a red herring last time, but we found the issue now! You have 2 Eclipse accounts, and the one that's linked to this GH account doesn't have an ECA, while the one that has an ECA doesn't have the Github account link. We'll need to at least clear the GH field of the unused account if you want to update the account you've been using, but the real solution is fixing the accounts and merging them into one. Can you check in on https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4780#note_2547110 so we can get that started for you?

autumnfound:
Also to clarify, I can confirm at this time that the email you used in association with the commits DOES have an ECA, but when we do Github checks, we use the Github username when available as it's an explicit link to accounts. That explains why when we check your email its fine, and why your account has an ECA, but fails here.

aminissn:
> Also to clarify, I can confirm at this time that the email you used in association with the commits DOES have an ECA, but when we do Github checks, we use the Github username when available as it's an explicit link to accounts. That explains why when we check your email its fine, and why your account has an ECA, but fails here.

Is there any specific solution for this? I am using the same email address for both Github and Eclipse.

aminissn:
can we merge these changes now? so I can keep working from the latest development version of the main distribution for bus priority?

m-kro:
As @behrisch wrote, please avoid the f-strings for compatibility reasons. Something more:

- We don't use tabs for indentation in XML (output)
- Test [complex/traci/bugs/ticket4324](https://sumo.dlr.de/extractTest.php?path=complex/traci/bugs/ticket4324) fails - could you have a look why this is the case?

behrisch:
@aminissn I integrated the base of your work concerning earlyTarget and conditions. Most of the functions you added did not look immediately useful to a broader audience so I skipped them for now. I think there is no real benefit in wrapping all the phase access again if the user can already retrieve the phase list and work with it. The Condition class was not used so I skipped it as well. I hope with some minor adaptions the bus_priority script will still work. Please open another PR for that one! Thanks!

PR #15247, opened 2024-07-18
Add queue length calculation to tls for GLOSA device

Domsall:
Apart from the new function, I updated the queue length calculation:

- The calculation seemed to only take one lane into account (E2 detectors support multiple lanes since a while).
- And I changed some E2 detector parameters to the default values. This also changes a few tests in delay_based traffic lights. So please check, if you support this update.


m-kro:
Can you please look at the failing tests in sumo/tls/delay_based?

Domsall:
Yes, as I wrote, I changed the values of the E2 detector to the defined default values:

- timeThreshold was 0, now 1 s
- speedThreshold was 0, now 5/3.6 m/s
- jamThreshold was 0, now 10 m

m-kro:
Sorry I meant you could yourself as well have a look at the tests and commit the changes in case it seems reasonable.

Domsall:
From what I can tell, it was previously not working correctly (a bug). No vehicles were recognized on any lanes. Now vehicles/a jam is detected on the left turning lanes.

Domsall:
@m-kro Is something still missing or failing?

m-kro:
There is still a failing test (traci/get_variable/Vehicle/ID_COUNT_subseconds) but I don't see how it is connected to the pull request. We'll solve that one after accepting it...

Issue #14694, opened 2024-04-10
feat: add support for directly writing output to parquet files


mschrader15:
Not an issue, just a proposal.

I [recently implemented a simple parquet writer](https://github.com/mschrader15/sumo-pipelines/blob/sumo_cpp/cpp/include/libsumo_wrapper.h) using the [Arrow Parquet stream writer](https://github.com/apache/arrow/blob/main/cpp/examples/parquet/parquet_stream_api/stream_reader_writer.cc) and libsumocpp

Simulation w/ the parquet wrapped simulation seems to be slightly faster than native sumo writing to XML, but when you combine that w/ the subsequent read (using `sumolib.xml.parse_fast`) it is ~10x faster. This speed up likely grows w/ the length and size of simulation.

namdre:
we could maybe support this based on the user-defined file name extension similar to how automatic .xml.gz compression/decompression is implemented.

behrisch:
After some discussion with @namdre @aminissn @mschrader15 and @stprnvsh the guideline is as follows:

- We only support parquet for selected output files and improve bit by bit.
- Initially the user needs to provide the attributes which shall be written.
- Later we will add a "discovery" mode which allows to detect the needed attributes but only from the first "full depth" XML write.
- Warnings shall be present if unknown attributes are being written
- Buffering larger writes to detect all needed attributes is not in scope
- At some future point schema support can be added
- We start with supporting FCD output
- If there are changes needed in writing our outputs we will support that (for instance to allow type detection).
- Meandata is probably next.

behrisch:
I just added a CSVFormatter in ae3d73538aaef73deadeff149c4a205772ee9c13 to showcase how a different formatter might be implemented without touching the streams (and also because it might be useful anyway). Maybe the parquet implementation can follow these lines?

mschrader15:
I think it can, loosely. The issue is that `parquet::StreamWriter` does not inherit from `std::ostream`, which means the Parquet implementation can't be quite as tightly integrated.

behrisch:
@mschrader15 @stprnvsh There is now a first draft in the parquet branch. Please have a look if we can continue from there

behrisch:
Some preliminary numbers with writing default fcd output for the acosta scenario (which needs 8s without fcd output on my machine) and reading back with the pandas.read_x function. I also added sumolib.xml.parse_fast_nested for comparison

format|write|size | pandas.read | parse_fast_nested
[...]

PR #15311, opened 2024-08-02
parquet support for structured output

mschrader15:
Quickly drafted an implementation of parquet using the OutputDevice base type. I had to abstract out the dependence on `std::ostream` into a `StreamDevice` to get it to work somewhat nicely. 

Using templating + polymorphism (and mimicking the behavior of `std::ostream`) tested my c++ skills, and I'm not happy with heavy use of static_cast. 

Wanted to open this in draft to see if the SUMO team has any immediate feedback / suggested design changes before I clean it up and add in tests?

Current Issues:

- [ ] It currently expects all attributes to have appeared by the first time `closeTag` is called. This works for outputs like emissions export, but won't for every output type.
     - [ ] This also fails when (in the emissions export case) a vehicle is not present in the first timestep. We need to add a device-specific expected depth or similar.
     - [ ] We need a runtime check for compatibility or to use more advanced features of parquet like [Dremel encoding](https://blog.x.com/engineering/en_us/a/2013/dremel-made-simple-with-parquet)
- [ ] Arrow/Parquet requires cpp17
- [ ] Build w/o `HAS_PARQUET` defined fails per the pipelines


namdre:
Thanks for getting started on this! 
Summer holidays are coming around so expect the review to be delayed until september :sweat_smile:

behrisch:
@mschrader15 Thanks for this one! It already looks quite advanced. Maybe we should split it into two parts and do the unique_ptr refactoring first? Also is c++17 really required for using parquet?

mschrader15:
Re the C++17, unfortunately yes. See https://github.com/apache/arrow/issues/32415

FYI, I put together an informal speed comparison here: https://github.com/mschrader15/sumo-parquet-test?tab=readme-ov-file#results

I have a busy next month, but will try to find time to break this apart. Agree that it is probably best as several smaller PRs for review-ability.

mschrader15:
I am going to revist this. To be clear @behrisch, you want this in two MRs?

1. Refactor of existing output functionality + std::ptr functionality
2. Addition of parquet support

aminissn:
can we help to push this PR? I think protobuf for I/O would be very helpful to speed up outputs parsing for large networks

mschrader15:
Hey @aminissn the current blocker is the requirement of C++17. 

I'm willing to push ahead on this once we identify whether it is possible to continue

behrisch:
I think we can go forward here but I would like to keep C++17 part a little bit isolated if possible. We need to check for the presence of parquet anyway, so if we can "ifdef" this I am just fine. I would still like to have this in two different PRs though. One for the pointer refactoring, one for parquet.

behrisch:
Another thing which I noticed is that structurally Parquet should only be a formatter but not a separate device because in the end it should not matter whether I write the parquet into a file or the network.

build/src/config.h
behrisch, line 1: Please do not add config.h. It is a generated file

namdre:
@behrisch: I was just reminded by @aminissn how useful this would be. Can we do something on our end to integrate this feature? After all, @mschrader15 may not have the time to work on it.

aminissn:
on our side @stprnvsh has been continuing the initial work of @mschrader15. We have a working build at: https://github.com/Transcality/sumo-transcality

behrisch:
@aminissn @stprnvsh Wow, there is already a lot of new code in your repo. I would definitely like to merge this but do you think  we can somehow do this step by step because you touch many sensitive parts? Nevertheless I would like to start rather sooner than later because otherwise there will be a lot of merging effort I suppose.

aminissn:
@behrisch maybe we can have a short meeting and discuss where to start from? You are totally right, we have done many changes and would like to merge them carefully and systematically.

mschrader15:
Hey all, still following this. 

I closed the MR because my merge of [0063a63](https://github.com/eclipse-sumo/sumo/pull/15311/commits/0063a63300cc718fb3759c7724d0406158308ee2) unintentionally brought in a bunch of changes. 

I was working locally on a cleaner solution, but lost the time to finish it. Will push where it is in the next couple of days. It will bring in all the relevant changes that @stprnvsh made, while also limiting scope to parquet writing.

behrisch:
Hi @mschrader15 ,
great that you are still onboard! I has a talk with @stprnvsh and @aminissn last week and we wanted to start with a minimal working example which only covers FCD output. It will be published in the [parquet](https://github.com/eclipse-sumo/sumo/tree/parquet) branch in this repo. Maybe we can find a way to synchronize our work? We can talk here or I can open up a channel at our mattermost or matrix?

mschrader15:
Cool! I just pushed a (admittedly simple) commit to https://github.com/eclipse-sumo/sumo/pull/16782 that build a uniform interface for Parquet Stream Writer and std::ostream's

Next step is a cleaned up ParquetDevice and ParquetFormatter. I will push in the coming days. 

If that isn't the direction that you are going to go (I think there is ultimately a larger structural change needed for parquet output of more complicated outputs), just let me know. Happy to let you take this over, just wanted to get all my local changes cleaned up and pushed for reference!

behrisch:
I was still hoping that we can get away without touching the streams themselves and just change/add formatter and device. Do you think that is feasible? I would rather not add another encapsulation. If it is necessary I would prefer adapting the (fcd-)output itself.

mschrader15:
yea, I think thats possible and the right way to proceed given the scope laid out in #14694

behrisch:
@mschrader15 @aminissn @stprnvsh @namdre I propose to continue the discussion in #14694

PR #15330, opened 2024-08-08
Fix Misc.md

sean-reed:
Updated ptLine attributes table with correct name, default, and description for nightService.


m-kro:
Thanks

PR #15341, opened 2024-08-12
Retrieve lane index for vehicle in meso sim

xmskk:
Changed getLaneIndex(veh_id) in Vehicle.cpp to allow retrieval of lane index for meso sim vehicles (as long as meso-lane-queue is set to true)


namdre:
Thanks!

Issue #14681, opened 2024-04-09
retrieve segment index for vehicle


namdre:
requested in #14670

Issue #14670, opened 2024-04-06
Best way to color meso segments/edges


zihaooo:
Is there a way to render segments in given colors in SUMO meso? It would be great if I could set the color of segments via Traci. 

If there is no way to color segments, how can I color edges?

I found this issue https://github.com/eclipse-sumo/sumo/issues/9462 mentioned there is a file that can be used as input to set the color of edges.
I also found this issue https://github.com/eclipse-sumo/sumo/issues/9842 mentioned there is a Traci API that can change the edge color.

Could you please give the basic examples to demonstrate how to use these two methods?

Thank you for your help.

namdre:
For xml-data based coloring, please try to follow the documentation at https://sumo.dlr.de/docs/sumo-gui.html#visualizing_edge-related_data
For traci-coloring, set the coloring scheme to color edges "by param" and use `traci.edge.setParameter`
See also https://sumo.dlr.de/docs/sumo-gui.html#gui-settings_files for how to work with coloring schemes.

Feel free to post another question here if you struggle with the documentation.

zihaooo:
Thank you very much for your answer. 
In SUMO meso, the edge is cut into segments based on `meso-edgelength`. Since my target is color segments, is there any way I can get the cut segments' geometry, or can I know which segment a vehicle is on?

zihaooo:
In SUMO (meso) GUI, I can right-click a vehicle and select the `show parameter` in the right-click menu. In the new pop-up window, I can see it is on segment 2 of the current edge (which can be shown in the attached screenshot). 
![Screenshot from 2024-04-08 17-19-05](https://github.com/eclipse-sumo/sumo/assets/64977999/a0515391-14cd-47fa-a0b1-490ac744c5ef)

How can I get this **segment** info from traci? If I need use `traci.vehicle.getParameter` to retrieve this parameter, what is the name of this parameter? I have already tried `traci.vehicle.getParameter(veh_id, 'segmentID')`, etc. which has empty return

namdre:
There isn't yet a way to access a vehicles segment or the segment geometry. TraCI was built with microsim in mind so the meso features are lagging behind, sorry.

zihaooo:
Thank you so much for your reply. 

> For traci-coloring, set the coloring scheme to color edges "by param" and use `traci.edge.setParameter`
> See also https://sumo.dlr.de/docs/sumo-gui.html#gui-settings_files for how to work with coloring schemes.

[...]

PR #15342, opened 2024-08-13
Retrieve segment id and segment index from vehicle

zihaooo:
I added two functions to retrieve segment ID and segment index. 
Tested with both `libsumo` and `traci` via Python.
In the mesoscopic simulation, calling these two functions will return the corresponding segment info (e.g. `"EG_23_2_1:1"` and `1`.
In the microscopic simulation, calling these two functions will return the empty string `""` and `INVALID_INT_VALUE` (-1073741824).
Refs #14681 and  #14670.

It's my first time to create a pull request. I follow all the steps here: https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work.

I am not sure about the following two things:
1. How do I choose a new constant in `src/libsumo/TraCIConstants.h`? Currently, I use `0xa1` and `0xa2` for `VAR_SEGMENT_ID` and `VAR_SEGMENT_INDEX`, respectively.
2. Is `tools/traci/_vehicle.py` generated from `src/libtraci/Vehicle.cpp` or it can be manually edited? Currently, I manually edit it based on what I've done in `src/libtraci/Vehicle.cpp`.

Please let me know how I can improve this pull request.


namdre:
Thanks for the contribution!

1. It's a bit of a dark art. We mainly try to ensure that constants can be re-used between domains. Since 0xa1 is only used for VAR_VIEW_OFFSET (the CMD_GET doesn't matter), and since there is probably no use for retrieving something like a view offset from a vehicle, it's fine to re-purpose the constant for segment id retrieval.

2. _vehicle.py is maintained manually rather than generated

The only thing missing before I can close #14681 is a test but I shall add this myself after merging.

PR #15581, opened 2024-10-11
Update osmGet.py

5lood237:
Added a small print that shows which tile it is at currently,


behrisch:
Thanks for the inspiration! I took the liberty to adapt the code a little and the paperwork seems to be missing as well, see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work, so I am closing this one

PR #15614, opened 2024-10-18
GLOSA Device Update

Domsall:
Apart from many new Debug output commands, we added more options to the GLOSA device:

1. New option `device.glosa.use-queue` that uses an E2-Detector to calculate the queue length of vehicles in front of the traffic light, which is used to update the arrival time of the GLOSA-vehicle
2. We updated the calculation of the next traffic light's phase. The GLOSA-vehicle now not only calculates the time to the next phase switch, but to the phase when the vehicle arrives at the intersection (phase calculation is in a loop).
3. Added a switch time offset to the looped phase calculation. This makes GLOSA-vehicles drive a bit faster when they want to reach the beginning or end of a green phase. Result: Fewer stops and higher traffic flow. 
4. Updated one minor test change


m-kro:
@Domsall Some meta tests still fail (probably due to the new option). Can you please run them and eventually confirm the new outcome (if you see nothing else wrong with them)?

Domsall:
you are right, I did not know that the options can be written to an output and that tests exist for that. The new queue option is throwing the error, everything else is fine.

src/microsim/devices/MSDevice_GLOSA.cpp
```diff
@@ -270,7 +278,18 @@ MSDevice_GLOSA::notifyMove(SUMOTrafficObject& /*tObject*/, double oldPos,
                 }
             }
             // calculate next Phase
-            nextSwitch += getTimeToNextSwitch(myNextTLSLink, currentPhaseGreen, currentPhaseStop, countOld); 
+            nextSwitch += getTimeToNextSwitch(myNextTLSLink, currentPhaseGreen, currentPhaseStop, countOld);
+            // For vehicles far away from the junction we add an offset
+            if (nextSwitch > 20. && switchOffset == 0) {
+                nextSwitch -= (double)switchOffset;
```
m-kro, line 284: It does not seem useful to me to subtract 0?

src/microsim/devices/MSDevice_GLOSA.cpp
```diff
@@ -118,37 +124,149 @@ MSDevice_GLOSA::notifyMove(SUMOTrafficObject& /*tObject*/, double oldPos,
...
+#ifdef DEBUG_QUEUE
+                if (DEBUG_COND) {
+                    std::cout << SIMTIME << " veh=" << myVeh.getID() << " Nextswitch(RedQueue)=" << nextSwitch << "\n";
+                }
+#endif
+            }
+        }
+
+        // Search for the next passable phase, maximum 10 Phases 
+        for (int countwhile = 1; countwhile < 10; countwhile++) {
```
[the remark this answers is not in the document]
Domsall: You are right :) You could also use an "endless" while-loop, we wanted to have a safety-net

src/microsim/devices/MSDevice_GLOSA.cpp
```diff
@@ -118,37 +124,149 @@ MSDevice_GLOSA::notifyMove(SUMOTrafficObject& /*tObject*/, double oldPos,
...
                     }
+#endif
+                }
+                adaptSpeed(myDistance, timeToJunction, nextSwitch + myAddSwitchTime, solved);
+                if (countwhile == 2 && myUseQueue) {
+                    // reset queue because following phases should not calculate with current queue
+                    nextSwitch -= QueueLength * factor + addition;
+                    timeToJunction -= additionaljunctiontime;
+                }
+                if (solved == true) {
```
m-kro, line 276: no need for the equal operator as the variable is already a boolean

src/microsim/devices/MSDevice_GLOSA.cpp
```diff
@@ -118,37 +124,149 @@ MSDevice_GLOSA::notifyMove(SUMOTrafficObject& /*tObject*/, double oldPos,
...
-                if (myMaxSpeedFactor > myVeh.getChosenSpeedFactor()) {
-                    const double vMax2 = vMax / myVeh.getChosenSpeedFactor() * myMaxSpeedFactor;
-                    const double timetoJunction2 = earliest_arrival(myDistance, vMax2);
-                    // reaching the signal at yellow might be sufficient
-                    const double yellowSlack = myVeh.getVehicleType().getParameter().getJMParam(SUMO_ATTR_JM_DRIVE_AFTER_YELLOW_TIME, 0);
+        if (myUseQueue) {
+		    // Detect queue length at tls
+		    double QueueLength = myNextTLSLink->getTLLogic()->getTLQueueLength(myNextTLSLink->getLaneBefore()->getID());
+            double greentime = 0;
+            double additionaljunctiontime = 0;
```
m-kro: camel case variable naming style is preferred, see https://sumo.dlr.de/docs/Developer/CodeStyle.html#variable_names

Domsall: Should I change it or will you?

src/microsim/devices/MSDevice_GLOSA.cpp
Domsall: Should I change it or will you?

src/microsim/output/MSE2Collector.cpp
```diff
@@ -1553,15 +1553,15 @@ double
 MSE2Collector::getEstimateQueueLength() const {
 
     if (myVehicleInfos.empty()) {
-        return -1;
+        return 0;
```
m-kro, line 1556: Although this code part does not seem to be used much, it is an additional value to know immediately whether the queue length is due to absence of vehicles or they just didn't stop...

src/microsim/devices/MSDevice_GLOSA.cpp
m-kro: You can do it

src/microsim/devices/MSDevice_GLOSA.cpp
m-kro: You can do it

Domsall:
All comments are fixed

namdre:
@Domsall It appears that the new option `device.glosa.use-queue` only takes effect for `tlType="delay_based"`.  Do you have some results on how often the dynamic switching invalidates the GLOSA-advisory?

Domsall:
> @Domsall It appears that the new option `device.glosa.use-queue` only takes effect for `tlType="delay_based"`. Do you have some results on how often the dynamic switching invalidates the GLOSA-advisory?

Yes... but we only used the `delay_based` type for the creation of the E2-detector. The traffic lights ran in a static loop.

Domsall:
> also, would you care to add info on the new features at https://sumo.dlr.de/docs/Simulation/GLOSA.html ?

Its on my agenda

PR #16007, opened 2025-01-14
#15861 Update documentation on dockerized SUMO

### LINKED ISSUE #15861: Update Developer Documentation for Dockerized SUMO
The [Docker section of the developer documentation](https://sumo.dlr.de/docs/Developer/Docker.html) is currently severely outdated, since it does not mention the Docker images being uploaded to the GitHub-Container-Registry. (The last time that page was updated was one day after the initial commit to the Dockerfiles).

I would suggest updating that page to refer to the official Docker Images (and maybe even add some examples for using them).

**SUMO-version:** Current website

**operating system:** -

Menkalian:
I updated the developer documentation on Docker and added a tutorial on the basic usage in docker (and fixed a unsatsified link error occurring in the container).

I tried to get an example with a GUI working, but I did not get it to work reproducible.

I hope I managed to hit the right tone for the docs.

closes #15861

### DIFF
--- a/build_config/docker/Dockerfile.ubuntu.git
+++ b/build_config/docker/Dockerfile.ubuntu.git
@@ -30,6 +30,7 @@ RUN cd /opt; git clone -b $JUPEDSIM_VERSION https://github.com/PedestrianDynamic
 RUN cd /opt; cmake -B jupedsim-build -D CMAKE_INSTALL_PREFIX=jupedsim-install jupedsim; \
  cmake --build jupedsim-build -j4 --config Release; cmake --install jupedsim-build --config Release; rm -rf jupedsim-build
 RUN echo "/opt/jupedsim-install/lib" > /etc/ld.so.conf.d/jupedsim.conf
+RUN ldconfig
 
 # python packages needed for the tools (the ones where we do not have ubuntu packages)
 RUN pip3 install -r /opt/sumo/tools/requirements.txt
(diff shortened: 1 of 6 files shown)


Menkalian:
I found a small issue in the docker image, that prevented sumo from executing. The new commit fixes that.

behrisch:
@Menkalian Thanks a lot! Instead of mangling with the ld path, I think it might be easier to install jupedsim directly into the system since we are dockerized anyway. What do you think?

Menkalian:
> @Menkalian Thanks a lot! Instead of mangling with the ld path, I think it might be easier to install jupedsim directly into the system since we are dockerized anyway. What do you think?

Yeah, that's probably more the "docker" way. I was not sure if there was a specific reason for keeping a separate install directory, but it might just be a remnant from some previous work.

I will take a look at it.

If I find any further improvements for the docker image (e.g. smaller size, adding some meta information, checking with best practices, ...), would you prefer for me to add it in this PR or open a separate issue and PR? I'm fine with either way.

Menkalian:
Hey there,
Just wanted to ask whether there is something missing from my side for review/merge of this PR. I know some of the actions failed, but as far as I could tell, these were not caused by me.
Please give me a hint, if I have to fix them or add anything to this PR. Thanks 🙂 !

behrisch:
Thanks a lot! This will help many people. I was a little unsure whether the tutorial should refer to "bleeding edge" main. Maybe it would be better to use "latest" here. What do you think?

Menkalian:
The reason I chose `main` instead of `latest` was, that `latest` did not work when I tested it (the jupedsim path problem) and I wanted the tutorial to work as soon as it was merged.
Generally speaking, I agree that using the more stable version `latest` should be the preferred way. Maybe that could be updated when a new version of `latest` is available.

PR #16132, opened 2025-02-05
Docker tutorial: Refer to latest stable image instead of bleeding-edge

Menkalian:
As suggested by/discussed with @behrisch in #16007 

Now the latest image works with the tutorial and therefore the stable image is preferrable to the potentially unstable one.

relates to #15861


behrisch:
Thanks a lot! And we would very much appreciate if you find the time for a docker GUI tutorial as well :-)

Menkalian:
Yeah, sure! I had to attempt a few times, but now I got it working. Expect a PR soon 😉!

PR #16579, opened 2025-04-22
added vType to generateTurnRatios.py

aminissn:
I tried to add vType to the turn-ratio output so one can get turns per passenger, truck, bicycle etc.

### DIFF
--- a/tools/turn-defs/generateTurnRatios.py
+++ b/tools/turn-defs/generateTurnRatios.py
@@ -77,16 +77,17 @@ def getFlows(options):
     maxDepart = 0
     begin = 0
     interval = options.interval if options.interval is not None else 1e20
-    # intervalBegin -> edge -> followerEdge -> count
-    intervalEdgePairFlowsMap = defaultdict(dict)
+    # intervalBegin -> edge -> followerEdge -> veh_type -> count
+    intervalEdgePairFlowsMap = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: defaultdict(int))))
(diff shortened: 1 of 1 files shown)


namdre:
Looks useful.

1. I propose this be made optional (--split-types)
2. I think it would be more useful to write only one edgeRelation with multiple attributes of the form `<typeID>_<count>`. Other tools such as routeSampler.py, turnCount2EdgeCount.py can select the attribute for reading counts but they cannot perform filtering based on some other attribute such as `vType`

aminissn:
i agree with 1, but on point 2, then the probability calculations would be confusing. Would it be the ration of bikes over cars on an incoming road, or like as it is implemented now, ration of bikes in different directions? if we keep them separate, it would be "easier" to explain, no?

namdre:
The semantics would stay as they are: the values of an attribute are only compared to the same attribute in other edgeRelations.
I agree that it needs some explaining but the big advantage is, that it's useful out of the box and doesn't require rewriting other tools.  Also, by default, the values aren't even turning probabilities but simply absolute turning counts and this makes them quite easy to interpret.

aminissn:
I somehow have an issue again with eclispe eca, but I think their server has issues because I cannot login. Yesterday I renewed the eca and the test was passed, today it is failing again

aminissn:
yes, parsing speed could become important, we model large networks for longer period so the xml output can be relatively large. Is there a reason why not to use `parse_fast_nested` in general? Because of the name and order of attributes?

PR #16815, opened 2025-06-17
fixed unclosed element in closingReroute XML

davide990:
The "closingReroute" tags were missing the closing '/' char. I fixed both script and the tests.


namdre:
Thanks for spotting this!

namdre:
[ECA checked](https://github.com/eclipse-sumo/sumo/pull/16751#issuecomment-2939453736)

PR #16825, opened 2025-06-20
Fix for 'shift' option in traceExporter.py script

JelenaT30:
'_Shift_' option within traceExporter.py script only offsets the **_negative_** _x_ and _y_ coordinates and leaves **_positive_** _x_ and _y_ coordinates untouched. This leads to the fact that whole trace is messed up after "shift". 

In case '_shift_' option is used **_all coordinates_** shall be shifted, not only negative ones.


namdre:
Thanks!

PR #16901, opened 2025-07-15
meso fcd-ouput interpolate like gui

aminissn:
I tried to "improve" the fcd output in meso to offer a lower resolution of positions similar to the way that vehicles are interpolated in the gui. 
I hope it is the right way without causing issues.


behrisch:
Thank you for the PR! I think it is okay if GUI and FCD output deviate but as soon as we touch one output, we need to make sure that the others report the same (especially also with TraCI / libsumo). This looks like a bigger effort to me which does not mean we should not do it. But maybe we should start a branch here. @namdre What do you think?

behrisch:
The code should be made reusable also for TraCI, please do not duplicate code. @aminissn We can start a new branch together if you like

aminissn:
> The code should be made reusable also for TraCI, please do not duplicate code. @aminissn We can start a new branch together if you like

happy to contribute, given my limited c++ skill

behrisch:
we want to integrate this

aminissn:
> we want to integrate this

Amazing. I tested this change, the visualization of fcd-output looks very smooth now. But I agree, this should be done for traci/libsumo as well. Any other parts of the code that need to be adapted as well?

aminissn:
I accidentally force pushed this pr. I create a new one with the changes..

PR #17081, opened 2025-09-15
device.rerouting only when improvement above a threshold

aminissn:
could this be a solution for https://github.com/eclipse-sumo/sumo/issues/4635 ?
I do observe more stable results after all the discussion in https://github.com/eclipse-sumo/sumo/issues/17058


namdre:
Thanks. It's a bit more complicated than that because there are other sources of rerouting that should not be affected (i.e. after adding an new stop). I'm already working on it.

namdre:
resolved via 2e22dc737a56aa5ad1e485dcd80cfc9bc1c11c39

namdre:
if you have a use case where `previousCost` is negative, please share it.

PR #17191, opened 2025-10-08
Meso-fcd output position interpolate in all outputs

aminissn:
A new PR that I hope is a complete and correct version for https://github.com/eclipse-sumo/sumo/pull/16901


aminissn:
can I push this PR @behrisch? If modifications are needed I am happy to do so.

behrisch:
I finally managed to have a look. We still have several issues here:

1. The GUI calculation is still different because it takes into account the position of the downstream vehicle to avoid vehicle positions being not increasing. If we want to do that we probably need caching of vehicle positions because we do not want to recalculate the whole edge for every vehicle. 
2. Angle and (longitudinal) lane position are not in sync with the x,y pos after the change. We should try to avoid this. While the angle is probably  not so relevant there is even a remark in the code that interpolating the lane position can cause problems.
3. Do we want to keep the old behavior available optionally?
4. The unrelated speedRelative should not be part of this PR.

behrisch:
I started a branch here: https://github.com/eclipse-sumo/sumo/tree/refs/heads/interpolated_meso

behrisch:
I deleted the branch and pushed my first version to main. It is still hidden behind an option (--meso-interpolate-pos) because I did not check thoroughly but already saw non-monotonic positions which is not nice. But please test @aminissn

behrisch:
the basic work is done, everything else continues in #17945

PR #17312, opened 2025-11-10
Added generated Czech language translation files.

MSpatenka:


m-kro:
Does that mean you actually want to translate UI messages into Czech as well? Generating the templates only makes sense if they get filled to a reasonable degree.

MSpatenka:
Yes, I generated them in order to start translating. Sorry for not making that intention clear.

PR #17328, opened 2025-11-13
add/NOx-noise-to-fcd

aminissn:
If I am not wrong the fcd-ouput supports emission outputs as well,. but somehow NOx and Noise were not included. I added these missing attributes and modified the docs accordingly.


namdre:
Thanks!

PR #17384, opened 2025-12-03
read xml-attributes from xsd file for sumolib.parsing

aminissn:
Does this actually help with generalizing the sumolib.xml.parse() family function? 
This was part of a discussion earlier in the mailing list about adding overlapDensity in the schema file. 
Currently, this function allows me to read xml files like below without worrying about the order of the attributes: 
`for row in sumolib.xml.parse_fast_nested(input_xmlfile, 'interval', ['begin'], 'edge', sumolib.xml.attributes_from_xsd(f'{$SUMO_HOME}/data/xsd/types/meandata.xsd', 'edgeLaneDataType'), optional=True):`

### DIFF
--- a/tools/sumolib/xml.py
+++ b/tools/sumolib/xml.py
@@ -385,7 +385,90 @@ def _get_compound_object(node, element_types, element_name, element_attrs, attr_
         [attr_conversions.get(a, _IDENTITY)(node.get(a)) for a in attrnames],
         child_dict, node.text, child_list)
 
+def attributes_from_xsd(xsd_path, type_or_element_name):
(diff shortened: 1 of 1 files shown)


behrisch:
Thanks for this one! Making parsing easier is definitely a big plus. This code however relies on the assumption that the order in the xsd is identical to the output order. We can try to keep an eye on this but currently there is no check enforcing it. Furthermore it is not clear to me why the code cares so much about namespaces since we do not use them.

behrisch:
I fixed this by using the existing xsd parser

PR #17467, opened 2025-12-24
Fix VRAM memory leak in decal texture handling and make viewport / decal list persistence work as intended 

wpenistone:
Fixes VRAM memory leak caused by decal textures not being properly cleaned via GL. (+ fonts [about 4mb])

Properly persists viewport and decal list upon reload (ctrl+r), 
Initially it appeared that the decal list & viewport was meant to clear on reload, but looking deeper at the code, it seems that they were intended to be kept / persist and I have made changes to ensure this happens as previously intended. 

Testing via netedit;
Before;
image

After;
image


wpenistone:
PS: first time getting into SUMO development, works for me and looks elegant enough, let me know of any issues.

namdre:
Thanks for contributing!

It seems you are trying to fix two issues at once:
1) memory leak
2) viewport/decal persistence 

If feasible, please do a separate pull request for each one.

Also for us to check the issue and their solution please provide a list of steps to replicate the memory leak.
Does this affect only netedit or also sumo-gui?

For 2) whic only affects netedit, I'd like to hear about your use case for using reload where you'd like to preserve viewport and decals.
Whereas sumo-gui has two obvious use-cases for reload (starting the simulation again, loading modified inputs) and sumo-gui already perists viewport and decals, the use case for reloading in netedit is less obvious and thus the expected behavior is also less clear.

wpenistone:
> Thanks for contributing!
> 
> It seems you are trying to fix two issues at once:
> 
> 1. memory leak
> 2. viewport/decal persistence
> 
> If feasible, please do a separate pull request for each one.
> 
> Also for us to check the issue and their solution please provide a list of steps to replicate the memory leak. Does this affect only netedit or also sumo-gui?
> 
> For 2) whic only affects netedit, I'd like to hear about your use case for using reload where you'd like to preserve viewport and decals. Whereas sumo-gui has two obvious use-cases for reload (starting the simulation again, loading modified inputs) and sumo-gui already perists viewport and decals, the use case for reloading in netedit is less obvious and thus the expected behaviour is also less clear.

Yes, I was first after the vram leak before also noticing that, not ideal, feel free to separate it; not available for a couple days.

It affects both, simply load a decal list and clear list / reload / load another network, it will not properly clear video memory. 
This is especially noticeable if you have large decals (see below), often times you're forced to restart the application.
image


It's usually expected to be a normal UX thing for reloads within applications, e.g. specifically for netedit: hot loading network & imagery modifications.
And to be consistent with sumo-gui.

wpenistone:
Are the failing checks normal? If preferred, we can close this one and I'll separate the changes into two PRs :) 
 Done, See https://github.com/eclipse-sumo/sumo/pull/17606 & https://github.com/eclipse-sumo/sumo/pull/17607

PR #17473, opened 2025-12-29
netconvert: Improve --heightmap.geotiff

opatut:
This pull request adds a few improvements for people who want to use the heightmap feature to generated 3D coordinates:

* add `--offset.z` to the value read from the heightmap input (otherwise it is unused, so this way you can offset the input data properly)
* use float instead of integers when reading a heightmap (gdal converts this for us)
* read the spatial reference from the geotiff file and use it for projection, so you don't have to manually (and lossily) convert the file first by reprojection (e.g. `gdal raster warp`)

I've also thrown in a change that uses [bilinear interpolation](https://en.wikipedia.org/wiki/Bilinear_interpolation) instead of the previous triangle logic for interpolation of the raster data. It seems to produce the same results (rounded to centimeters) on my data and is probably faster to execute, and simpler to understand :) You can skip that commit if you like, it is refactoring only and adds no user benefit, but I prefer the cleaner logic and simpler math.

### DIFF
--- a/src/netbuild/NBHeightMapper.cpp
+++ b/src/netbuild/NBHeightMapper.cpp
@@ -92,10 +92,28 @@ NBHeightMapper::getZ(const Position& geo) const {
         const Boundary& boundary = item.boundary;
         float* raster = item.raster;
         double result = -1e6;
-        if (boundary.around2D(geo)) {
+
+        double x = geo.x();
+        double y = geo.y(); 
+
+#ifdef HAVE_GDAL
+        // Transform geo coordinates to the coordinate system of this
+        // raster image for lookup in its raster, if applicable.
+        if (item.transform != nullptr) {
+            // Since the input coordinates are always WGS84 (they may be
+            // transformed to it in NBNetBuilder::transformCoordinate), and
+            // WGS84 uses latitude-longitude order (y-x), we have to swap the
+            // input coordinates here.
+            std::swap(x, y);
+
+            item.transform->Transform(1, &x, &y);
+        }
+#endif
+
+        if (boundary.around2D(x, y)) {
             const int xSize = item.xSize;
-            const double normX = (geo.x() - boundary.xmin()) / mySizeOfPixel.x();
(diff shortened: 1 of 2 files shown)


namdre:
Thanks! Can you check whether your ECA is in order? (https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work)

opatut:
There was an issue while filling out the ECA, I've contacted the licensing team to resolve it. Will post here once it's done.

opatut:
It seems like I didn't test it well, and there is an issue with my bilinear interpolation change, producing incorrect heights. I've removed that commit from this PR and will work on it some more, independently. The rest still works fine.

behrisch:
Thanks! @opatut If you have an example we can add to our tests, it would be more than welcome.

opatut:
I've been using heightmap files from the German state of Baden-Wurttemberg's state office of geo information, which are licenced [rather openly](https://www.govdata.de/dl-de/by-2-0), so you might be able to integrate them:

https://opengeodata.lgl-bw.de/#/(sidenav:product/dgm1)

They come in [EPSG:25832 (UTM zone 32N)](https://epsg.io/25832) by default, so projection is necessary.

Combine this with an OpenStreetMap excerpt, and you got your test dataset. This is from my Makefile:

```Makefile
netconvert \
  --verbose \
  --sidewalks.guess \
  --crossings.guess \
  --heightmap.geotiff $(HEIGHTMAP) \
  --offset.x -$(ORIGIN_X) \
  --offset.y -$(ORIGIN_Y) \
  --offset.z -$(ORIGIN_Z) \
  --proj.utm=true \
  --geometry.max-segment-length=4 \
  --ptstop-output=$(OUTPUT)/osm_stops.add.xml \
  --ptline-clean-up \
  --ptline-output=$(OUTPUT)/osm_ptlines.xml \
  --railway.topology.repair \
  --output-file $(NET_FILE) \
  --osm-files $(OSM_FILE) \
  --keep-edges.in-geo-boundary $(BOUNDS)
```

PR #17579, opened 2026-01-23
Import PT Modes in osmWebWizard

### LINKED ISSUE #8628: ferry import in osmWebWizard
May need an extra option or could be always done if public transport is enabled

ozsoyisik:
Implementation:
- Added checkboxes for all public transport modes to the UI below the main 'public transport' checkbox.
- Converted the checkbox values to additional ptOptions for calling ptlines2flows.main
- Added the ships-typemap to netconvertOptions if the ferry checkbox is active (it should be disabled by default).

Behavior:
- If the top-level Import Public Transport checkbox is enabled, netconvert always exports PT stops and ptlines (--ptstop-output and --ptline-output) regardless of which pt checkboxes are selected.
- The pt checkboxes affect only ptlines2flows (via --types), in oder words the public transportation flows; they do not prevent netconvert from exporting stops/ptlines.
- If no modes are selected, ptlines2flows is not called

Note: I was a bit lost about how to fix tests so I wanted to push my code to see the current test situation (On my local, there were many tests failing)

Close #8628


ozsoyisik:
Hey, is it possible to run the workflow? @namdre

ozsoyisik:
@namdre As I said earlier, I struggled to understand the tests because in some of the tests I was not sure if I need to change the numbers for example. Also, is it possible for you to review my code beforehead so that I can at least fix my code first and then check the tests? Or can you provide me guidance in a way that you think appropriate?

namdre:
Thank you for your contribution! (and sorry that it took me so long to review it).
Texttest requires a mocked out `data` dictionary and this needed to have the new ptModes dictionary for tests to succeed ([b7bb97a](https://github.com/eclipse-sumo/sumo/commit/b7bb97a3d76c9fb22993a9993ae9f97a41b05a56))

ozsoyisik:
Thank you for your patience and support through the process! It was my first experience with open source. I am so happy that my contribution has been merged. 😄

PR #17590, opened 2026-01-29
Missing installation commands in the test setup document

### LINKED ISSUE #17578: Tests Documentation Suggestion
While following the test document, I realized that TkDiffv is not directly opened to display the differences. However, I got two errors:
1. Cannot find file viewing program '**tkdiff**'.
Please install it somewhere on your PATH or
change the configuration entry 'view_program'.
2. Cannot find file viewing program '**emacs**'.
Please install it somewhere on your PATH or
change the configuration entry 'view_program'.

Then I run those two to install those packages:
brew install tkdiff
brew install emacs

I believe it is better if those commands are also given in the documentation. If it is okay, you can assign this issue to me. Thanks.

ozsoyisik:
Documentation improvements:

* Added instructions to install recommended diff viewer (`tkdiff`) and text editor (`emacs`) using Homebrew, as well as alternatives and configuration tips, in `docs/web/docs/Developer/Tests.md`.

This pull request adds instructions to the developer testing documentation to help users set up recommended tools for viewing test results on macOS. The most important change is the addition of guidance for installing a diff viewer and text editor. With that addition, it will be easier for beginners to run the tests. (Based on my experience:)) 

Closes #17578

### DIFF
--- a/docs/web/docs/Developer/Tests.md
+++ b/docs/web/docs/Developer/Tests.md
@@ -65,7 +65,6 @@ sudo apt-get install meld geany
 **On Windows:**
 
 Download and install from
-[Meld project page](https://meldmerge.org/) and
 [Geany project page](https://www.geany.org/) by following the provided installation instructions.
 
 After installing these tools, you need to configure TextTest to use them by creating or editing `$HOME/.texttest/config` (see [Customize configuration](#customize-configuration) section below).


ozsoyisik:
@behrisch The current version covers changing the TextTest defaults in the "Customize configuration" section, which shows how to set diff_program and view_program in $HOME/.texttest/config. And in my contribution, I referred to that to make it clearer. If this is not enough, I can make the relationship more explicit.

behrisch:
Thanks for the contribution! I think it is not a good idea to let people first install tkdiff and emacs just to tell them later that we actually prefer meld and geany. tkdiff and emacs are not our recommendation, they are just the texttest defaults. So unless _you_ really prefer emacs and tkdiff I think we should not mention them early on. We should rather tell people directly to install geany and meld and change their config (and only mention emacs and tkdiff as fallback). What do you think? Maybe there is also a good macOS default editor we could mention here?

behrisch:
And we should clarify that you need to install the tools you refer to not only on macOS. The page should cover all OS.

ozsoyisik:
Okay but I did not understand one point. You said that you (sumo developers) prefer meld and geany. However, there is a line saying that:
"Double-clicking on the red box *battery.sumo* opens ''TkDiffv''"
then why tkdiff is specified here? Is this a mistake or done on purpose?

ozsoyisik:
Hello I applied my changes based on your comments. I also checked for better alternatives but I could not find better alternatives. @behrisch

behrisch:
Thanks, it looks a lot better now! Concerning your questions: I think the tkdiff reference was a leftover and probably different SUMO developers (and doc authors) do have different preferences here. So tkdiff is very basic and very fast but it cannot even handle the clipboard properly, so I would definitely not recommend it. But probably people who get that far into the SUMO docs already have their favourite editor / diff tool already. Meld even gets installed on Windows together with texttest  so there is no need to mention it. Feel free to adapt to that, if not it is fine as well and I will merge it soon.

ozsoyisik:
I have removed the Meld command from the Windows part as you suggested. Thanks for your help throughout the process! @behrisch

behrisch:
Thank you!

PR #17601, opened 2026-02-04
Added ETCS rail following model

edoaxyz:
The current rail following model generates safe speeds by relying on a fixed deceleration which is not applicable in most railway cases: usually the braking capabilities depends on the current speed, so the sum of forces is made from the (electrical) braking force, the air resistance force, and the forces applied by slope gradients. In a real context, resistance forces due to turns should be considered too, but the required effort is above my time limits.

The model is named `RailETCS` because this is how braking curves are computed by EVC systems (European Vital Computer) in ETCS equipped trains, even though this is a simplified version of the whole process. The safe speed is computed always in the same method which can:
- integrate all the partial distances traveled during each speed step, for each that goes from the current speed to the target one, using the acceleration of the mentioned forces, being careful to shift the position in which the slope gradient is picked;
- use a precomputed map of braking distances made only of resistance and braking forces, and later add the slope forces so that the outcome is less accurate (rounded to the next higher value) but it is computed in constant time, by using a rounding method for both speed and distances, so that there are a number of speed steps equal to the number of values in the `speedTable`, and a number of distance steps defined by the *vType* variable `numSamples`.

I've added also some tests just to check that sumo works with this model in both fixed and moving block signalling.

Also I've decided to not integrate this changes in the `Rail` model as it will introduce breaking changes for users who rely on that model, and also because the implementation is still not complete at all, given that not all forces are considered and also mass factor of trains has been neglected.

### DIFF
--- a/src/microsim/cfmodels/MSCFModel_RailETCS.cpp
+++ b/src/microsim/cfmodels/MSCFModel_RailETCS.cpp
@@ -189,7 +189,7 @@ double MSCFModel_RailETCS::minNextSpeedEmergency(double speed, const MSVehicle *
 double MSCFModel_RailETCS::getSafeDecel(const MSVehicle *const veh, double gap) const
 {
     ETCSVehicleVariables *vehVars = dynamic_cast<ETCSVehicleVariables *>(veh->getCarFollowVariables());
-    return vehVars->getMinDecl(veh, gap) + GRAVITY * sin(DEG2RAD(vehVars->getMinGradient(veh, gap)));
+    return myTrainParams.getBraking(veh->getSpeed()) / myTrainParams.weight + GRAVITY * sin(DEG2RAD(vehVars->getMinGradient(veh, gap)));
 }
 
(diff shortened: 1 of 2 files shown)


namdre:
Looks great! If you think the new model is already valuable in it's incomplete form, I'm happy to merge it.
Community value would benefit from documentation added at https://sumo.dlr.de/docs/Simulation/Railways.html#modelling_trains 

A question regarding your tests: in your vType you set `numSamples="900000"` and this causes vehicles to compute a lookup table with ~4e11 entries. This makes the tests quite slow. Was this intentional or would a lower value still lead to useful model behavior?.

edoaxyz:
Unfortunately I'm closing this PR as the model has been refactored and at the end it looks like the existing rail following model that considers only variable deceleration. Also as you might have noticed, the initial computation of braking maps is quite heavy and still needs some improvements.

namdre:
Consider adding a new parameter to the existing rail model which activates the ECTS deceleration computation. This avoids code duplication and achieves backward compatibility.

PR #17607, opened 2026-02-07
Persist viewport and decal list on reload in netedit

wpenistone:
Updates netedit to persist the current viewport and decal list during a network reload (Ctrl+R). aligns the behaviour with sumo-gui


namdre:
Thanks!

PR #17621, opened 2026-02-11
Change phase duration type from `int` to `float` in sumolib to avoid `ValueError` when loading net with `sumolib`

zihaooo:
We can set the float phase duration in NetEdit. So, it should be loaded as a float with sumolib.net.readNet to avoid `ValueError: invalid literal for int() with base 10: ...` 
Also, this change is consistent with the comment in https://github.com/eclipse-sumo/sumo/blob/1f4b28c24cacacdcbc6fc3324dfe5e53c21a27b3/tools/sumolib/net/__init__.py#L109

### DIFF
--- a/docs/web/docs/Simulation/Traffic_Lights.md
+++ b/docs/web/docs/Simulation/Traffic_Lights.md
@@ -172,15 +172,14 @@ The following attributes/elements are used within the tlLogic element:
 
 Each phase is defined using the following attributes:
 
-| Attribute Name | Value Type            | Description                |
-| -------------- | --------------------- | -------------------------- |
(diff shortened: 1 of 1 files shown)


namdre:
sumo does work with subsecond tls timing and it's okay to permit this in sumolib.
However, since it's a rare use case, it makes sense to still use ints if possible even in sumolib, thereby shortening outputs.

Furthermore, netconvert does not support floats for options pertaining to tls timings and sumo-gui does not (yet) support displaying fractional tls timings. For this reason we do not yet wish to advertise these capabilities in the documentation.
(Also, for consistencies sake it's not enough to change the documentation on `duration` but also `minDur,maxDur and `offset` are affected. 
See also #17622

PR #17809, opened 2026-03-30
Improve Overpass API download reliability in osmGet.py

josephmjustin:
## Problem

The current retry logic in `osmGet.py` only retries on HTTP 504 errors. Other temporary server errors (429 rate limit, 502 bad gateway, 503 service unavailable) cause the server to be removed from the URL list, exhausting available servers prematurely. Additionally, servers that return HTTP 200 with HTML error pages instead of OSM XML data are not detected, causing downstream build failures.

### osmGet.py
- Retry on 429/502/503/504 with adaptive backoff (exponential for rate limits, linear for server errors)
- Add `overpass.private.coffee` as a third default fallback server alongside `hpi` and `oapi`
- Validate that server responses contain `<osm` tag before saving — servers returning HTML error pages with 200 status are now detected and skipped
- Always print error details (server URL, HTTP code, retry delay) for better debugging
- Always print which server the download succeeded from
- Base delay is configurable via existing `--retry-delay` CLI option

### osmWebWizard.py
- Disable redundant subprocess call that triggered a second download as a side effect of config saving

## Testing

Tested against live Overpass servers with intermittent 504/502 failures. Verified:
- Retry logic correctly retries temporary errors without removing servers
- HTML error pages from servers are detected and the server is skipped
- Scenario builds successfully end-to-end through osmWebWizard

### DIFF
--- a/tools/osmGet.py
+++ b/tools/osmGet.py
@@ -47,21 +47,6 @@ TYPEMAP_DIR = os.path.join(THIS_DIR, "..", "data", "typemap")
 
 
 def readCompressed(options, urls, context, query, roadTypesJSON, getShapes, filename):
-
-    # initiate delay function for retrying download in case of specific errors
-    base_delay = options.retry_delay
-
-    def retry_delay(error_code, idx):
-        if error_code == 429:
-            delay = base_delay * 2 ** idx
-        elif error_code in (502, 504):
-            delay = base_delay + idx
-        elif error_code == 503:
-            delay = base_delay * 2 * (idx + 1)
-        else:
-            delay = base_delay
-        return delay
-
     # generate query string for each road-type category
     queryStringNode = []
 
@@ -178,17 +163,15 @@ def readCompressed(options, urls, context, query, roadTypesJSON, getShapes, file
         try:
             req = Request(url, data=finalQuery.encode(), headers=headers)
             with urlopen(req, context=context) as response:
+                if options.verbose:
+                    print(response.status, response.reason)
                 if response.getheader('Content-Encoding') == 'gzip':
                     lines = gzip.decompress(response.read())
                 else:
                     lines = response.read()
                 # validate before injecting SUMO header
                 if b"<osm" not in lines[:1000]:
-                    print("Removing %s from list of valid servers (returned non-OSM content)." % url, flush=True)
-                    urls = [u for u in urls if u != url]
-                    if urls:
-                        continue
-                    raise RuntimeError("All servers returned non-OSM content after %d attempts." % (idx + 1))
+                    raise ValueError("Server %s returned non-OSM content" % url)
                 declClose = lines.find(b'>') + 1
                 lines = (lines[:declClose]
                          + b"\n"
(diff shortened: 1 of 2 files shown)


behrisch:
Thanks a lot for this one! Some comments form my side:
I think the part with the adaptive delay is a little overengineered. Did it solve a real world problem or can we keep the old code?
Are you sure config saving triggered another download?
It was intentional that osmGet is running silently if there are no errors and verbose is not set.
Can you give an example of a server which returns 200 with an error page?

josephmjustin:
Thanks for the review. Addressed all points:

**1. Adaptive delay**
Agreed it was overengineered. Simplified to use `options.retry_delay` for all retryable errors (429/502/503/504). Removed the custom delay function.

**2. Config saving subprocess**
Yes - the subprocess runs osmGet.py with the same download args plus -C, so get() executes a full download as a side effect of saving the config. Then osmGet.get(osmArgs) calls the same function in-process, downloading again and overwriting the file. But I've reverted osmWebWizard.py to keep this PR focused on osmGet.py only.

**3. Silent by default**
Understood. Kept retry and success messages behind `options.verbose`, matching existing behaviour.

**4. Server returning 200 with error page**
Observed this with `overpass.private.coffee`. The server returned HTTP 200 but the body was an HTML error page instead of OSM XML:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
    "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
<head>
  <title>OSM3S Response</title>
</head>
<body>
<p>The data included in this document is from www.openstreetmap.org.</p>
<p><strong style="color:#FF0000">Error</strong>: runtime error: open64: 0 Success
/osm3s_osm_base Dispatcher_Client::request_read_and_idx::protocol_error</p>
</body>
</html>
```

This is an Overpass API behaviour when the server is overloaded — not specific to private.coffee. The `<osm` content validation catches this regardless of which server returns it.

If adding `pcoffee` as a third default server is a concern, happy to drop it and keep this PR focused on retry logic and content validation only.

behrisch:
Thanks for the quick adaptions. There are still unnecessary whitespace changes and the print for the status got lost

behrisch:
> **2. Config saving subprocess** Yes - the subprocess runs osmGet.py with the same download args plus -C, so get() executes a full download as a side effect of saving the config. Then osmGet.get(osmArgs) calls the same function in-process, downloading again and overwriting the file. But I've reverted osmWebWizard.py to keep this PR focused on osmGet.py only.

No, "-C" triggers a sys.exit. I agree this is bad style but it does not trigger another download

josephmjustin:
Fixed - removed whitespace change and restored the status print. Thanks for the clarification on -C, understood it exits before the download runs.

behrisch:
Can you just raise ValueError or catch RuntimeError to avoid duplicating the error handling? Also please re-add the newline at the end of the file

josephmjustin:
Done- replaced with raise ValueError to reuse existing handler, re-added trailing newline.

behrisch:
Thanks!

josephmjustin:
Thanks for the review and the merge.

PR #17829, opened 2026-04-06
free decal CPU image after GPU upload to avoid double memory usage

wpenistone:
Decal textures were kept in CPU RAM (via decal.image) after being uploaded to GPU VRAM, causing 2x memory usage per decals lifetime.
The CPU copy was never used after upload, so it has been dead code ever since (including the clean-up of it).

Delete the FXImage immediately after glTexImage2D upload, matching the pattern already used by GUITexturesHelper::getTextureID().


namdre:
Thanks!

PR #17913, opened 2026-04-23
Update gtfs2pt to handle trolleybus entries correctly and honor GTFS stop ids

jprk:
This PR updates the `gtfs2pt` script suite to handle `--mode trolleybus`, as the corresponding GTFS mode ids were missing. The underlying vClass for trolleybuses is `bus` and this has been reflected in some places as well.

The PR also replaces the numeric `line` attribute that has been copied from vehicle id with true public transport line ID to reflect what documentation says about using the attribute when planning person trips.

The final change is the optional possibility to replace the numeric stop ids in the format `<number>.<number>` with ids reflecting the GTFS stop ids. We are using this in our custom GTFS workflow to further modify the stop ids to be unique alphanumerical representation of the stop name (i.e. we are able to replace `gtfs_<id>` with `borska_pole#0` which simplifies the later analysis a lot).  This change required that the FCD vehicle output has been extended with an undocumented attribute `gtfsid`.


namdre:
Happy about the PR and happy to see you are still using SUMO (-:
for the record: The eclise ECA/check shows a failure but I verified the email in the patch and there is a valid ECA on file at eclipse.org.

namdre:
In [18ca1ab](https://github.com/eclipse-sumo/sumo/commit/18ca1aba535443b50572177cb90d562522d389ea) I'm reverting the route that was changed in addition to the line ID and I reverted the route. I'm curious why the behavior would have been different on your machine (it's the same for me on windows and linux and our test servers). If you happen to track down the reason, let me know.

namdre:
I've also moved the test to our internal test suite because they are too large and long-running for our "normal" test schedule. If you have a smaller test case (i.e. only one trolleybus line a 2-5 stations in a smaller subnetwork) I'd be happy to add it to the normal test schedule.

namdre:
> The final change is the optional possibility to replace the numeric stop ids in the format <number>.<number> with ids reflecting the GTFS stop ids. We are using this in our custom GTFS workflow to further modify the stop ids to be unique alphanumerical representation of the stop name (i.e. we are able to replace gtfs_<id> with borska_pole#0 which simplifies the later analysis a lot). 

@jprk I'd be happy to integrate this further renaming feature as well!

namdre:
@jprk For reasons explained in #17934, the default is to have distinct line ids (of the form ORIG_ID#INDEX) for trips that serve different stop sequences. If you absolutely want the unmodified line ids, you can set option **--original-lines**

PR #17962, opened 2026-05-08
fix: replace 122 bare except clauses with except Exception

KeloYuan:
## Summary

Replaces **122** bare `except:` clauses with `except Exception:` across **52 files** in `tools/contributed/hybridPY/` and `tools/net/`.

## Why

Bare `except:` catches `SystemExit` and `KeyboardInterrupt`, which can mask Ctrl+C handling and make debugging harder.


behrisch:
@KeloYuan Thanks for this one! Can you give an example where this was a problem for you? And can you double check whether you accepted the contributor agreement (see https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work)

behrisch:
I am closing this one for lack of feedback. We might come back to the suggestions especially concerning the hybridpy code later

PR #18023, opened 2026-05-20
fix/nan-fcd-parquet

aminissn:


aminissn:
@behrisch based on your comment earlier today, I added an option so that user can decide to write empty rows in the output to have a behavior similar to xml output writer

behrisch:
Great, but I think this option would be useful for the standard fcd output too, so no need to limit it to parquet

aminissn:
> Great, but I think this option would be useful for the standard fcd output too, so no need to limit it to parquet

I wasn't sure if you like that part of the code to change, due to backward compatibility or parsing functions. But if you think it won't cause any issues, I made the changes in my last commit.

behrisch:
I had a hard time trying to reproduce the initial error (division by 0). @aminissn Do you have an example for that?

aminissn:
> I had a hard time trying to reproduce the initial error (division by 0). @aminissn Do you have an example for that?

I can't reproduce it either. I think I got confused, since my vehicle departure time was 28802.39 and in the fcd output I saw 28802 none and only from 28003 it had values. Which is basically the second issue where empty rows are written.

behrisch:
I implemented the functionality for skipping empty timesteps independently (see the commits above)

PR #18213, opened 2026-07-16
Calculate queue length in meso

aminissn:
I tried to exploit Fable 5 for this one. Hopefully it is a working solution for: https://github.com/eclipse-sumo/sumo/issues/18212

### DIFF
--- a/data/xsd/queue_file.xsd
+++ b/data/xsd/queue_file.xsd
@@ -5,9 +5,10 @@
 
     <xsd:element name="queue-export">
         <xsd:complexType>
-            <xsd:sequence>
+            <xsd:choice>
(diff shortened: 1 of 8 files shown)


namdre:
Looking good.
I suggest to make the 95-percentile value configurable and keep the number out of the attribute name (i.e. renamed to `percentileQueueLengthInVehicles` etc)

aminissn:
> Looking good. I suggest to make the 95-percentile value configurable and keep the number out of the attribute name (i.e. renamed to `percentileQueueLengthInVehicles` etc)

I made the changes. I am actually looking forward to see the outcome of this combined with meso-lift.

PR #18235, opened 2026-08-13
Add NaSch (Nagel-Schreckenberg) cellular automaton car-following model

### LINKED ISSUE #12182: add carFollowModel NaSch (Nagel Schreckenberg)
pretty similar to a discretized version of Krauss 
- replace leaderSpeed with gap
- round returned speeds to multiples of cell size 
- accel = cellSize
- do not bound emergency decel

ideas for dealing with cell discretizaition
- round vehicle positions to multiples of the cell size
- either treat roads that are shorter than the cell size as if they had a single cell (meso style) - requires ignoring some  collisions
- or reprocess the network to ensure that all roads have at least the length of 1 cell

jerrylin118:
## Summary
- Implements the NaSch cellular automaton car-following model as `carFollowModel="NaSch"`
- Speeds are rounded down to multiples of the cell size implied by the vehicle's `accel` attribute (one cell gained per simulation step, per the issue's "accel = cellSize")
- The safe speed depends only on the gap to the leader/obstacle, not the leader's speed ("replace leaderSpeed with gap")
- Emergency deceleration is intentionally left unbounded, so a vehicle can lock onto its leader's speed within a single step ("do not bound emergency decel")
- `sigma` is reused as the per-step randomized-deceleration ("dawdle") probability

## Testing
- Added `tests/sumo/extended/carfollowing/nasch/basic`: a leader cruising at 15 m/s and a faster follower cruising at 30 m/s on a straight motorway. The follower accelerates in cell increments, catches up, and settles into the leader's speed.
- Built locally with CMake + vcpkg (MSVC 2022, Release) and ran the scenario to verify the trajectory and confirm the build is clean.

Fixes #12182


namdre:
Thanks for the contribution. This is looking good. Please take a look at the legal niceties (ECA) that must be covered before merging a pull request: https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work

namdre:
Thanks. Even though the on github check fails, I confirmed that everything is in order by checking https://accounts.eclipse.org/users/jerrylin118

PR #18250, opened 2026-08-21
Feature/cch routing

stprnvsh:
### Motivation

In large scenarios with `device.rerouting`, repeated shortest-path queries dominate
runtime: every reroute pays a full A*/Dijkstra search over the network. Customizable
Contraction Hierarchies invert that cost structure: a one-time metric-independent
hierarchy construction at startup, a cheap re-customization at each weight-adaptation
barrier, and microsecond point-to-point queries in between. CCH answers queries in ~0.01-0.02 ms exploring ~30 edges on average

### Design

- **Graph mapping** (`CCHGraph`): RoutingKit nodes are the non-internal, non-TAZ edges;
  arcs are the allowed edge-to-edge connections, with internal/via chains folded into
  the arc weight (`w(u→to) = viaEffort(u..to) + effort(to)`, centiseconds). The
  topology is metric-independent and built once; weights are refilled from the live
  effort function at each customization, so the arc set never changes.
- **TAZ connectors are deliberately not graph nodes**: eliminating a degree-d star
  creates a d-clique and poisons the hierarchy (treewidth argument; measured fill and
  query blow-up in README_CCH.md). Instead TAZ endpoints are resolved at query time as
  multi-source/multi-target sets (the "phantom node" pattern), using RoutingKit's
  native multi-source query.
- **Per-vehicle-class metrics over one shared topology**: each class present in the
  demand gets its own metric; arcs the class may not use (connection-level permissions,
  matching `getViaSuccessors(vClass)`) are `inf_weight`. Runtime closures are
  permission changes, so they flow into the metric the same way, with no query-time
  special-casing.
- **Double-buffered lock-free publish**: the main thread customizes the back buffer at
  the adaptation barrier and publishes with a release store; rerouting worker threads
  read the front metric with acquire loads. No locks on the query hot path.
- **Sparse re-customization**: instead of a full per-barrier rebuild, edges whose
  smoothed speed actually moved are marked dirty and only their arcs are recomputed and
  propagated (RoutingKit partial customization). A deadband
  (`--device.rerouting.cch-update-threshold.factor/.constant`) keeps moving-average
  jitter from marking thousands of arcs per tick; closures bypass the deadband. The
  customize is also skipped entirely when no query consumed the published metric since
  the last barrier. Setting `SUMO_CCH_VALIDATE=1` validates every partial customization
  against a full rebuild (used throughout our testing; zero divergence).
- **Exact A* fallback stays embedded** for the cases a shared metric cannot express:
  a class without a metric, unmappable endpoints, or per-query prohibitions that are
  not live permission closures (same reasoning as the existing CHRouter). Route costs
  match the A* reference; 
- **duarouter** gains the same backend (`RODUACCHGraph`).

### Scope and guarding

- Fully **opt-in**: CMake option `SUMO_WITH_CCH=ON` plus `ROUTINGKIT_ROOT=<path>`
  (`HAVE_ROUTINGKIT`). Built without it, SUMO compiles and behaves exactly as before.
- RoutingKit (BSD-2) is an **external dependency, not vendored**, keeping the source
  tree EPL-clean.
- Unsupported effort modifiers are rejected with a clear error rather than silently
  mis-routing: `weights.random-factor != 1`, routing preferences, and
  `device.rerouting.bike-speeds` (all per-vehicle, inexpressible in a shared metric).
  `weights.priority-factor` is supported; it is per-edge and baked into the metric.


namdre:
Thanks! Did you already do the Eclipse ECA paperwork? See https://sumo.dlr.de/docs/FAQ.html#how_do_code_contributions_work

stprnvsh:
I just signed it now

namdre:
the commit contains the email address "141918471+stprnvsh@users.noreply.github.com" which I don't think is valid. See https://patch-diff.githubusercontent.com/raw/eclipse-sumo/sumo/pull/18250.patch

stprnvsh:
fixed it now :)

CMakeLists.txt
behrisch, line 1: I would rather not include this. It seems very specific to your setup. If you want we can add a warning if the include dirs of arrow and parquet somehow differ

src/netimport/NIImporter_ArcView.cpp
behrisch, line 1: This is a separate issue. I just fixed it in the main branch

behrisch:
@stprnvsh I would prefer to include routingkit in src/foreign since it is not that big. This way it will always be present. What do you think? Also is it a lot of work to separate the openmp part? I would prefer to start just with the single-threaded routing

stprnvsh:
> @stprnvsh I would prefer to include routingkit in src/foreign since it is not that big. This way it will always be present. What do you think? Also is it a lot of work to separate the openmp part? I would prefer to start just with the single-threaded routing

Works fine for me, threading doesn't add too much speed anyway, can separate the openmp and keep it single threaded

behrisch:
I added a cch branch now where routingkit is already included. @stprnvsh can you redirect the PR (without openmp) to that branch?

stprnvsh:
> I added a cch branch now where routingkit is already included. @stprnvsh can you redirect the PR (without openmp) to that branch?

Sure will do that, will raise a pr there

stprnvsh:
Please let me know if any of the changes are not in the exact format or if you would prefer the code to be structured differently @behrisch

behrisch:
your changes are fine. Please remove the ifdef entirely and adapt the sumo meta tests such that the test suites can succeed.

behrisch:
CCH should also be added to the help string for the routing-algorithm option (this will again change the meta test results, so maybe do this before adapting the meta tests).

behrisch:
Is there any chance to merge / template the cchgraph instead of having two implementations?

stprnvsh:
> Is there any chance to merge / template the cchgraph instead of having two implementations?

yup can do this will commit the above changes as well

behrisch:
Even the remaining code of the two derived graph classes looks very similar. Can we completely merge here?

stprnvsh:
> Even the remaining code of the two derived graph classes looks very similar. Can we completely merge here?

they are slightly different due to how they synchronise but can put them together

behrisch:
I think it would be beneficial, all the cch specific routing code would live in utils/router then

behrisch:
I merged it into the branch and would now look at the duarouter tests running with cch and determine why they differ. I will get back to you then

stprnvsh:
> I merged it into the branch and would now look at the duarouter tests running with cch and determine why they differ. I will get back to you then


found a few issues running the test suite :

1. Legacy function="connector" edges were excluded from the CCH graph. 

2. Metrics ignored everything the effort reads from the vehicle because fills used veh=nullptr. 

3. Query source seeds also used nullptr

4. expandNodePath inserted junction-internal edges into returned routes,  would have broken simulation rerouting on nets with internal links. 

pushing fixes for this

PR #18270, opened 2026-08-27
CCH: fixes from running the duarouter test suite with the cch variant

stprnvsh:
Follow-up to #18250. I ran the duarouter suite with the cch variant and found three real bugs:

- Edges declared with function="connector" in the net were excluded from the hierarchy as if they were district connectors. They are normal routable edges for the exact routers, so TAZ entry/exit selection went wrong (trips/TAZ/taz_weights). Now only the paired "-source"/"-sink" district connectors stay out of the graph.
- The metrics were filled with a null reference vehicle, which ignored the type's maximum speed, the vClass specific speed limits and restriction-params (vclasses/vClass_specific_speeds, trips/restrictionParams). There is now one metric per vehicle type and the queries pass the vehicle.
- The path expansion re-inserted internal edges into the returned route. Routes never contain internal edges, so identical routes compared unequal (duplicate alternatives in the vclasses tests) and nets with internal links would have gotten invalid routes.

After the fixes I compared all 350 tests that write alternatives (4474 vehicles) against the dijkstra results: 4465 are identical in route and cost, 3 pick a different path with the same cost (the astar_landmark tests) and 6 differ in cost for known reasons: closingLane/one_straight matches what alts.duarouter.chwrapper already codifies for time-bounded closures, fifo and fifo_interpolate change weights below the weight-period (they are not in the chwrapper suite either) and trips/stops/jumps/jump is off by 0.01, the rounding unit of the metric.

What remains in the cch variant are the option rejections that the other variants also exclude (measure, random-factor, preferences) and the verbose/error output wording, which needs .cch reference files like chrouter and chwrapper have. I can generate those if you want.

### DIFF
--- a/src/duarouter/RODUACCHMetrics.cpp
+++ b/src/duarouter/RODUACCHMetrics.cpp
@@ -32,7 +32,6 @@ RODUACCHGraph::EffortOperation RODUACCHMetrics::myEffort = nullptr;
 SUMOTime RODUACCHMetrics::myBegin = 0;
 SUMOTime RODUACCHMetrics::myWeightPeriod = SUMOTime_MAX;
 std::map<RODUACCHMetrics::MetricKey, RODUACCHMetrics::ClassMetric> RODUACCHMetrics::myMetrics;
-std::map<std::vector<double>, int> RODUACCHMetrics::myProfiles;
 std::mutex RODUACCHMetrics::myLock;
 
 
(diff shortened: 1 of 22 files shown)


behrisch:
Yes, please add the .cch variants where needed. I also noticed that CCHGraphBase is not a base anymore so maybe we should rename it to CCHGraph

stprnvsh:
> Yes, please add the .cch variants where needed. I also noticed that CCHGraphBase is not a base anymore so maybe we should rename it to CCHGraph

have renamed and added the .cch variants :)

behrisch:
@stprnvsh Thanks! I think we now need to do the same with the sumo.cch tests

stprnvsh:
Ahh adding them now, sorry

stprnvsh:
https://github.com/eclipse-sumo/sumo/pull/18285

PR #18285, opened 2026-09-01
Fix/cch sumo suite

stprnvsh:
This ports the CCH routing support to the sumo.cch test suite as a follow-up to #18270, taking it from 144 failures to a single upstream-owned one (closingLane, whose base reference is also stale on the default suite). 

It fixes several CCHRouter fallback gaps (transient permission modes, closed endpoints, message handler propagation, connection mask invalidation after runtime permission changes), ports the per-type metric handling from duarouter to the simulation including a free-flow metric family behind MSNet::getRouterTT, and fixes a staleness bug where the lazy customization skip could serve rerouting queries a metric arbitrarily many adaptation barriers old

Metrics are now re-customized at every barrier where an effort moved once any metric exists, bounding staleness to one barrier like the CH weight period, with an epoch check that diverts to the exact fallback if that bound is ever violated. 

All metric machinery is centralized in a new utils/router/CCHMetricFamily.h with a static mode (duarouter, free-flow) and a live double-buffered mode (rerouting device), which shrinks the footprint in MSRoutingEngine below what the cch branch carried before and lets duarouter reuse the same code.

The remaining test differences are covered by cch reference files (frozen random-factor realizations matching the CHWrapper precedent, equal-cost tie-breaks, cheaper parking loops, and router statistics splits, with routingMode_ignore matching the behavior the astar variant references already accept) and testsuite exclusions for six tests of other routing algorithms, mirroring the duarouter variant. A new option device.rerouting.cch-ensemble (default 1, bit-identical to the previous behavior) additionally allows K frozen random-factor realizations per vehicle type with deterministic vehicle assignment, recovering the route diversity of per-vehicle randomization at K times the customization cost; duarouter.cch stays 672/672 and the default suite is unchanged throughout.


behrisch:
@stprnvsh Thanks! Now we should add the algorithm to marouter as well

PR #18301, opened 2026-09-11
randomTrips.py: add option to return to origin

### LINKED ISSUE #14817: randomTrips.py: add option to return to the origin (when using --intermediate)
this should ideally be the opposite-direction edge of the from edge

SomaFon:
Fixes #14817

Adds `--return-to-origin` for trips using `--intermediate`.

The destination is selected from the edge with reversed from/to junctions when available. If no such edge exists, the source edge is used as the destination.

The implementation does not use the `bidi` attribute. The tests use the existing `net3.net.xml` and `net.net.xml` fixtures.

`--return-to-origin` cannot be combined with `--jtrrouter`.

changelog: [`randomTrips.py`]: add `--return-to-origin`


namdre:
The code is mostly fine but you (or your AI companion) have misunderstood the concept of a bidi edge in sumo. This is not used for marking the return-direction but for the special case of modeling bidirectional driving on the same lane (or railroad track).  The return-edge must be discovered by topology (in the simplest case it is an edge with the reverse from-junction and to-junction of the origin edge). Also, if no return-edge exists, it is perfectly fine to return to the very source edge itself (i.e. in case of a divided road).   

I suggest the following:
- rewrite the code that selects the destination edge as explained above
- do not hallucinate test input networks with invalid properties, instead use the networks that already exist in tools/trip/randomTrips for the test cases

SomaFon:
Thanks, you're right. I had misunderstood the meaning of `bidi`.

I changed the destination selection to look for an edge with reversed from/to junctions, and to use the source edge when no such edge exists.

I also replaced the hand-written test network with the existing `net3.net.xml` and added a fallback case using the existing `net.net.xml`.

SomaFon:
The macOS build and one Linux wheel test timed out after 6 hours. Could someone rerun the failed jobs?

behrisch:
@SomaFon Thanks!

